PR #27, opened 2017-07-14
Fix mismatched links in README.md

djsiroky:
Links for ladybug.location and ladybug.psychrometrics were not correct


mostaphaRoudsari:
Thanks @djsiroky!

PR #37, opened 2018-02-02
added wea to and from json

AntoineDao:
Needed to fix this for annual study in honeybee radiance.

This fixes issue #38

PR #40, opened 2018-02-15
to and from json for wea

AntoineDao:
This pull is required in order for the annual grid based to and from json methods to work (https://github.com/ladybug-tools/honeybee/pull/204)


AntoineDao:
Sorry I just realised I duplicated #37... Will correct in the latter PR.

PR #43, opened 2018-03-11
WindRose Initial Commit

devngc:
The purpose of this commit is to receive comments on the development so far.
The windrose.py does all the wind data parsing. 
The geometry wrapper is being written as windroseplus.py
Also, this development is nearing 200 lines of changes. (That's the another reason why this PR is made)


ladybug/datacollection.py
```diff
@@ -406,7 +406,7 @@ def filter_by_hoys(self, hoys):
            filteredDBT = DBT.filter_by_hoys(hoys)
         """
         _moys = tuple(int(hour * 60) for hour in hoys)
-        return self.filter_by_moys(_moys)
+        return DataCollection(self.filter_by_moys(_moys))
```
mostaphaRoudsari: Maybe it is a better idea if `filter_by_moys` returns a DataCollection in the first place.

ladybug/datacollection.py
```diff
@@ -524,6 +524,24 @@ def average_monthly_for_each_hour(self):
         """
         return self.average_data_monthly_for_each_hour(self.values)
 
+    def data_to_HOY(self, condition="x>5"):
```
mostaphaRoudsari: `hoy` not `HOY`

ladybug/datacollection.py
```diff
@@ -524,6 +524,24 @@ def average_monthly_for_each_hour(self):
...
+        """
+        Provides a list of HOY for the datapoints in the
+        filtered data from a conditional statement.
+        """
+        if isinstance(condition, str):
+            pass
```
mostaphaRoudsari: If this is for type checking then you can be explicit about it. The check and pass is confusing. I suggest to do:

```
if not isinstance(condition, str):
   # rest of the code
```

or just use assert

```python
assert isinstance(condition, str), ValueError(error_message)
```

ladybug/datacollection.py
```diff
@@ -524,6 +524,24 @@ def average_monthly_for_each_hour(self):
...
+        else:
+            errorMessage = "The condition you set must be a string." +\
+                " One of the example is 'x>5'." + \
+                " It replaced the non string condition you entered." +\
+                " Please re-enter your condition as a string."
+            raise TypeError(errorMessage)
+            condition = "x>5"
```
mostaphaRoudsari: you're raising an error on line 539. You'll never get here.

ladybug/datacollection.py
```diff
@@ -524,6 +524,24 @@ def average_monthly_for_each_hour(self):
         """
         return self.average_data_monthly_for_each_hour(self.values)
 
+    def data_to_HOY(self, condition="x>5"):
+        """
+        Provides a list of HOY for the datapoints in the
```
mostaphaRoudsari: The method is returning datetime not hour of the year.

ladybug/datacollection.py
```diff
@@ -524,6 +524,24 @@ def average_monthly_for_each_hour(self):
...
+        filtered data from a conditional statement.
+        """
+        if isinstance(condition, str):
+            pass
+        else:
+            errorMessage = "The condition you set must be a string." +\
```
mostaphaRoudsari: should be `error_message`!
PEP8 👮

ladybug/windrose.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+    Draws the Ladybug windrose
+    Usage:
+    """
+    __slots__ = ("epw", "HOYs", "annualHourlyData",
+                 "windCondition", "dataCondition",
+                 "numOfDirections", "centerPoint",
+                 "scale", "maxFrequency")
+
+    def __init__(self,
```
mostaphaRoudsari: A couple of notes:
1. We have moved from `mixedCase` to `snake_case`. All input arguments and variables in code should follow this.
2. The __init__ method should be considered for the most obvious way that a user would create the class. For other cases we should create a `classmethod`. For windrose you can create one from two DataCollections for wind speed and wind direction. If you want to create one from an epw file then you can have a `from_epw` classmethod. See WEA class as an example.
3. Conditional statements should not be part of the arguments. You should add a method to handle that.
```python
wind_rose = WindRose(wind_speed, wind_direction)
filtered_wind_rose = wind_rose.filter_by_conditional_statement(statement)
```
```

File: ladybug/windrose.py [hunk 8]
```diff
@@ -0,0 +1,156 @@
...
+        Instantiates the Ladybug WindRose Object
+        """
+        self.epw = EPW(epw_filePath)
+        self.HOYs = HOYs
+        self.annualHourlyData = annualHourlyData
+        self.windCondition = self.check_condition(windCondition)
+        self.dataCondition = self.check_condition(dataCondition)
+
+    @staticmethod
+    def check_condition(condition):
```
mostaphaRoudsari: This can be removed as similar method is implemented for `DataCollection`. If you call DataCollection.filter_by_conditional_statement it will check_condition under the hood.

ladybug/windrose.py
```diff
@@ -0,0 +1,156 @@
...
+                if "x" in statement.lower():
+                    xcount += 1
+            if len(condition) == xcount:
+                return condition
+            else:
+                print ("All conditions shall be a valid string with" +
+                       "a variable 'x' in it.")
+                return None
+
+    def data_to_HOY(self):
```
mostaphaRoudsari: The same here. The point of using DataCollection class is not to rewrite methods like this.

ladybug/windrose.py
```diff
@@ -0,0 +1,156 @@
...
+            HOYList = []
+            for i in range(len(self.annualHourlyData)):
+                HOYList.append(
+                    self.annualHourlyData[i].data_to_HOY(
+                        self.dataCondition[i]))
+            HOYSets = [set(listItem) for listItem in HOYList]
+            HOYCommon = set.intersection(*HOYSets)
+            return sorted(HOYCommon)
+
+    def parse_wind_data(self):
```
mostaphaRoudsari: This method can be broken down into several properties. Let's think about the most common properties that one wants to get from a WindRose. Frequency for each direction? Average wind speed for each direction? Then each of them should be accessible directly. If you haven't watched this talk already I think you should watch it before revising the code here: https://www.youtube.com/watch?v=OSGv2VnC0go&t=9s

ladybug/datacollection.py
devngc: Just to share something I learned recently. In Python, the use of assertions shall be limited to the conditions in code that are not easily covered by the exceptions. With -o and -oo command line switches, the assert statements can be turned into null operations. Basis are [this](https://docs.python.org/3/using/cmdline.html#cmdoption-o) and [that](https://stackoverflow.com/questions/944592/best-practice-for-python-assert)

devngc:
Thanks a lot @mostaphaRoudsari for such an extensive review! I will surely look closely at your comments and make necessary changes.

PR #45, opened 2018-04-05
Honeybee Server initial test branch

AntoineDao:
This is the ladybug image I am using to build antoinedao/ladybug and then antoinedao/honeybee and used to test the honeybee server.


mostaphaRoudsari:
Thank you @AntoineDao! There is no changes here that we need to merge. I'm not sure why GitHub is confused about the changes in several files.

PR #47, opened 2018-05-26
Making Ladybug Python 3 compatable

ocni-dtu:
Proof of concept of making Ladybug Python 3 ready.

There was only a few things that needed to be changed:
- izip was changed to zip
- xrange was changed to range

To make it work the "future" module is needed on Python 2 (can be pip installed)
For Python 3, euclid3 should be used instead of euclid, which only works on Python 2

All the tests are passing both on Py2.7 and Py3.6, there isn't that many tests, so it could still break

### DIFF
--- a/ladybug/datacollection.py
+++ b/ladybug/datacollection.py
@@ -8,7 +8,7 @@ from builtins import range
 
 try:
     from itertools import izip as zip
-except ImportError: # will be 3.x series
+except ImportError:  # will be 3.x series
     pass
 
 
--- a/ladybug/datacollection.py
+++ b/ladybug/datacollection.py
@@ -3,7 +3,13 @@ from .header import Header
 from .datatype import DataPoint
 
 from collections import OrderedDict
-from itertools import izip
+
+from builtins import range
+
+try:
+    from itertools import izip as zip
+except ImportError: # will be 3.x series
+    pass
 
 
 class DataCollection(object):
@@ -71,7 +77,7 @@ class DataCollection(object):
     @classmethod
     def from_data_and_datetimes(cls, data, datetimes, header=None):
         """Create a list from data and dateteimes."""
-        _d = (DataPoint(v, d) for v, d in izip(data, datetimes))
+        _d = (DataPoint(v, d) for v, d in zip(data, datetimes))
         return cls(_d, header)
 
     @classmethod
@@ -129,7 +135,7 @@ class DataCollection(object):
         return sum(values) / len(data)
 
     @staticmethod
-    def group_data_by_month(data, month_range=xrange(1, 13)):
+    def group_data_by_month(data, month_range=range(1, 13)):
         """Return a dictionary of values where values are grouped for each month.
 
         Key values are between 1-12
@@ -151,7 +157,7 @@ class DataCollection(object):
 
         return hourly_data_by_month
 
-    def group_by_month(self, month_range=xrange(1, 13)):
+    def group_by_month(self, month_range=range(1, 13)):
         """
         Return a dictionary of values where values are grouped for each month.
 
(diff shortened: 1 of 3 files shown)


ocni-dtu:
I commented out the .flip() on line 783. It doesn't seem to have any effect on the vector? And it doesn't exist in euclid3.

mostaphaRoudsari:
@ocni-dtu, thank you so much for the PR. That was quick. I'm traveling this week. Will get to this next week after the workshops.

Is it possible to also map xrange to range similar to what you did with izip > zip?
http://python-future.org/compatible_idioms.html#xrange

ladybug/sunpath.py
```diff
@@ -780,7 +780,7 @@ def _calculate_sun_vector(self):
             .rotate_around(z_axis, self.azimuth_in_radians) \
             .rotate_around(z_axis, math.radians(-1 * self.north_angle))
 
-        _sun_vector.normalize().flip()
+        _sun_vector.normalize()  #.flip()
```
mostaphaRoudsari: This is actually important and changes the direction of vector. I might have added it myself. If it results in incompatibility we can do something like:

```python
 _sun_vector.normalize()
_sun_vector =  Vector3(-1 * _sun_vector.x, -1 * _sun_vector.y, -1 * _sun_vector.z)
```

ocni-dtu:
I have changed xrange to range as well. But maybe it should be inside a try statement like izip. "Update 1" is mainly about changing xrange.

I think the tests should be updated to make sure both python 3.x and python 2.7 works. I would suggest moving to pytest and using tox for the testing the different versions

ladybug/datacollection.py
```diff
@@ -3,7 +3,13 @@
 from .datatype import DataPoint
 
 from collections import OrderedDict
-from itertools import izip
+
+from builtins import range
```
mostaphaRoudsari: @ocni-dtu this will fail in Python 2 as there is no `builtins` module.

mostaphaRoudsari:
Hi @ocni-dtu, I left a couple of notes.

I agree that we need to right more tests before merging this in. I can see that we will have issues that are currently in the code. For instance we use `itertools.izip` several times in wea module: https://github.com/ladybug-tools/ladybug/blob/751d27adb2461c150bae020d63b04e67ddfdd1b6/ladybug/wea.py#L467

or there are a number of cases which we use iterkeys and iteritems for dictionaries such as this line: https://github.com/ladybug-tools/ladybug/blob/master/ladybug/epw.py#L1028

and I'm sure there are more cases like this that we don't see right now. Let's not merge this for now until we have better tests in place. I started strong on writing tests but then didn't write enough tests or any test in some cases. It's on me to get it fixed unless someone else writes the tests before I do.

Thanks again for this PR!

ladybug/datacollection.py
ocni-dtu: Correct Py2 doesn't ship with builtins it is a part of the future package. See this: http://python-future.org/compatible_idioms.html#setup

ocni-dtu:
I see a need for this swift to Py3 and I have a couple of projects where it would be benefitial, so I would gladly help pushing it forward (that is why I raised the issue in the first place). 

If we could make an issue or project here on Github, so we can organize the work and progress. Then I could see a possibility for collaboration.

mostaphaRoudsari:
@ocni-dtu, sounds good! I will start opening the issues this weekend.

mostaphaRoudsari:
@ocni-dtu now we are ready for your PR. Added python 3.6 to test and it is currently failing!
https://travis-ci.org/ladybug-tools/ladybug/jobs/401003572

ladybug/datacollection.py
mostaphaRoudsari: our challenge is to keep the code compatible with IronPython too. Otherwise we can't use the libraries directly from Grasshopper interface.

mostaphaRoudsari:
@ocni-dtu see #52. I'm going to close this PR in favor of #52 which addresses the same issue! Thank you for getting the process started.

PR #49, opened 2018-06-23
Added Travis test + coverage

AntoineDao:
Did a brief rejig of testing setup by adding tests.py and .travis.yml. Main difference is that I created a test suite in the tests.py which avoids issues with finding the right python 2.7 and running each test seperately. I also added coverage testing to the travis pipeline. This gives you an idea of how much of the code base is actually covered by the tests written. Finally I've also added badges (because they look cool) which look for the Travis and Coveralls.io testing info and auto-update.

Please don't merge this without testing and discussion: I notice that coverage is low and that the build fails (only one test fails at this stage and I don't feel fixing it is something that should be covered by this pull request).

TL;DR: added automated testing configuration for Travis-CI.


README.md
```diff
@@ -1,3 +1,6 @@
+[![Build Status](https://travis-ci.org/AntoineDao/ladybug.svg?branch=master)](https://travis-ci.org/AntoineDao/ladybug)
```
mostaphaRoudsari: @AntoineDao, can we put this under ladybug tools account for travis? e.g.: `https://travis-ci.org/*ladybug-tools*/ladybug`

mostaphaRoudsari:
looks good! thanks. 💯 

@saeranv, any comments on this PR?

saeranv:
@mostaphaRoudsari @AntoineDao 

Looks pretty good to me also! I haven't worked with coverage testing before, but I think it's a great idea to add it. Two questions for @AntoineDao:

1. Can you clarify what the test suite is doing? Isn't the Travis build already using the same python interpreter (as specified in the .travis.yaml) for all of the tests it finds?

2. Is it possible to do this using pytest? I believe we are trying to transition away from unittest to pytest.

README.md
AntoineDao: Yes absolutely. I left them under mine so you could see what the badges look like :)

AntoineDao: Yes absolutely. I left them under mine so you could see what the badges look like :)

AntoineDao:
Hi @saeranv! 

1. Testsuite is just a native unittest way to string multiple tests in one script and return the report + errors. I found this made it easier to run/manage Travis.

2. Yeah I'm happy to give it a go if you think it's best that way.

I also reckon I might use the `coverage` result/reports to write a few more tests so ladybug can start on good footing. 

Ps: my moan reason for doing this now is to facilitate the deployment of docker containers for ladybug and honeybee so we can run on cloud ;)

saeranv:
@AntoineDao cool, I can see the advantages of having all the tests centralized on the tests.py script. 

Other then that everything looks good. If you can do the pytest transition that's good, but I'll leave it to @mostaphaRoudsari to decide how important it is at this stage. 

Excited to see you guys get this in docker!

mostaphaRoudsari:
> Other than that everything looks good. If you can do the pytest transition that's good, but I'll leave it to @mostaphaRoudsari to decide how important it is at this stage.

Yes. pytest is the right direction. Based on the documentation pytest also supports unittest cases which means we can still use the current cases.

Our main challenge right now is to write more tests!

AntoineDao:
Is it cheating if I use this: https://github.com/pytest-dev/unittest2pytest ? Seems to work just fine on ladybug.

image

saeranv:
So the tool rewrites the actual source files? That seems pretty useful actually, There's a lot of unittests in honeybee+ that could benefit from this.

AntoineDao:
The tool keeps the unittest basis for it but rewrites the unittest assertions into pytest readable format. It ran in sub 3s for the whole ladybug test suite. I'll give it a go on honeybee this evening and report back how it goes.

mostaphaRoudsari:
Since we want to move to pytest and the majority of tests needs to be rewritten would you suggest to rewrite the tests? I was under the impression that pytest can handle both cases with no extra effort but based on what I can see here I was wrong.

saeranv:
Good point. I remember I made some really minor changes to the testing classes for pytest: https://github.com/saeranv/honeybee-1/commit/ab293e1a8e34a6ee0ed7d016f671a2edfb6789f2#diff-d0f94bbbee023c9763159a41834472c7 But to be honest, I'm not even sure that even those minor changes were necessary...

AntoineDao:
Righto! I've added a ton of tests (some of which are breaking, probably because they are poorly written) and dramatically increased the test coverage from 30% to ~65%. I have also converted all tests to pytest using that nifty script I mentioned earlier on in this thread.

This is just a beginning! The tests I wrote are ok but to be honest could do with more clever thinking (is it worth testing use cases / common recipes rather than doing I/O checking for eg?).

Would be good to have a quick call with a repo manager (looking at you @mostaphaRoudsari  or @chriswmackey ) to help setup travis + coveralls in order to get neat badges + test each commit. I'm also happy to discuss how we currently use travis at my new work to provide a working example.

mostaphaRoudsari:
@AntoineDao, Oh wow. Well done! Now that you have this started writing more tests will be much much easier. I'm available tomorrow for most of the day. Let me know your schedule and we can set up a call. You know where to find me! ;P

@chriswmackey and @saeranv let me know if you guys want to join the call.

tests/epw_test.py
```diff
@@ -18,11 +19,34 @@ def tearDown(self):
...
+        assert epw_rel.file_path == relative_path
+        assert epw_rel.location.city == 'Chicago Ohare Intl Ap'
+        assert epw.file_path == abs_path
+        assert epw.location.city == 'Chicago Ohare Intl Ap'
+        # Check that calling location getter only retrieves location
+        assert epw.is_data_loaded == False
+        epw.import_data()
+        assert epw.is_data_loaded == True
+
+    # Should these tests fail? I'm not sure I understand what the Tokyo.epw is
```
mostaphaRoudsari: It's a non-standard epw file generated by one of the users. It used to fail but doesn't fail anymore! It's part of the test to ensure we can handle non-standard epw weather files.

AntoineDao:
😢 build fails.

mostaphaRoudsari:
Here we go! Time to merge this and get to the next steps.

We need to add tests for Python 3 now and then get to #47

PR #52, opened 2018-07-07
Ladybug <3 Python 3!

mostaphaRoudsari:
This is minimum changes that was neccessary to make ladybug work both in Python 2 and Python 3. Now we need to write more and more tests. I'm sure there are methods there that will fail in Python 3.
Resolved #47

Co-Authored-By: Christian Kongsgaard
Co-Authored-By: Antoine Dao

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
   - "3.6"
 
 install:
- - pip install pytest pytest-cov
+ - pip install pytest pytest-cov euclid3
 
 script:
   - py.test --cov=. tests/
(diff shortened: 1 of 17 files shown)


chriswmackey:
Thank you for doing this @mostapharoudsari,
I see that you also cleaned up a couple of poor coding practices in the Stat class. Things generally look good as far as I can tell (with my limited Python3) experience. The only issue that I caught is that it seems there should be a 
xrange = range
at the top of the Stat class.
Good job!

chriswmackey:
ah wait. I see that you addressed it by changing xrange to range in the file. I guess because we're only iterating over 9 items, this is ok.

PR #55, opened 2018-07-07
Several Fixes for Timestep Calculations

chriswmackey:
The most controversial change here that requires @mostaphaRoudsari 's review is bullet point 3 below.

I spent a lot of time thinking about how we want to deal with the interpolation of values in the data collection (and whether they fall on the half hour or the hour).  At first, I tried doing the implementation we initially thought we would do, where I just check the data type in the header against a list of values that I know to be on the half hour instead of the hour (specifically, radiation and illuminance).  But, as I started to think about it more, I realized that all of the E+ and Radiance results that we want to turn into data collections would eventually have to be added to this list since all of them are better represented by the half-hour than the hour (and adding the hundreds of E+ outputs to this list is not practical).  So I considered doing it in reverse, where the data type in the header is checked against a list of types that I know to be on the hour.  But it seemed wrong that the default would be to interpret the data as being on the half-hour when all of the data in the collection had datetimes that were on the hour.  Ultimately, I  just thought that everything would be much simpler if this half_hour was just an attribute that can be assigned to the data collection.  This solves the issue of having two locations where the half_hour nature of the epw data types is being stored (both in the epw class and in the data collection class).  It also frees up the data collection to deal with both results from E+ and Radiance (which we know to be half-hourly) as well as actual weather data (which is almost always on the the hour unless it's EPW radiation or illuminance).

Still, I recognize that this is not the most initially obvious solution and so I want your review @mostaphaRoudsari .

To give a full summary of everything in this commit:
1) The tox.ini file has been edited such that it now abides by our rule of 89 characters per line
2) The methods for shifting epw data depending on whether the values occur on the hour or half-hour have been cleaned up and made more efficient.  I reasoned that simply using boolean value titled "half_hour" is better than a float "start_hour" since houlrly data only occurs on the half hour or the hour mark.
3) Possibly the most controversial of change is that I added a property to the data collection class to track whether values should be interpreted as falling on the half hour or the hour.  This was done in lieu of checking  the type of the data in the header with a list of half-hour data types to look out for.  I will explain why I think this property is necessary when I send the pull request.
4) I added an option to specify data as cumulative on the interpolate data method, which is critical for using this method with radiation since sometimes you want the average flux at a timestep while other times you want the amount of solar energy that fell over the course of the last timestep.
5) I edited the Wea.from_epw_file method to allow for the input of a timestep.

### DIFF
--- a/ladybug/datacollection.py
+++ b/ladybug/datacollection.py
@@ -14,11 +14,12 @@ except ImportError:
 class DataCollection(object):
     """A list of data with a header."""
 
-    __slots__ = ('_header', '_data')
+    __slots__ = ('_header', '_data', '_middle_hour')
 
-    def __init__(self, data=None, header=None):
+    def __init__(self, data=None, header=None, middle_hour=None):
         """Init class."""
         self.header = header
+        self.middle_hour = middle_hour
 
         if not data:
             data = []
@@ -55,7 +56,7 @@ class DataCollection(object):
 
     @classmethod
     def from_list(cls, lst, location=None, data_type=None, unit=None,
-                  analysis_period=None):
+                  analysis_period=None, middle_hour=None):
         """Create a data collection from a list.
 
         lst items can be DataPoint or other values.
@@ -73,18 +74,21 @@ class DataCollection(object):
             return cls.from_data_and_datetimes(lst, analysis_period.datetimes, header)
         else:
             data = tuple(DataPoint.from_data(d) for d in lst)
-            return cls(data, header)
+            return cls(data, header, middle_hour)
 
     @classmethod
-    def from_data_and_datetimes(cls, data, datetimes, header=None):
+    def from_data_and_datetimes(cls, data, datetimes, header=None,
+                                middle_hour=None):
         """Create a list from data and dateteimes."""
         _d = tuple(DataPoint(v, d) for v, d in zip(data, datetimes))
-        return cls(_d, header)
+        return cls(_d, header, middle_hour)
 
     @classmethod
-    def from_data_and_analysis_period(cls, data, analysis_period, header=None):
+    def from_data_and_analysis_period(cls, data, analysis_period,
+                                      header=None, middle_hour=None):
         """Create a list from data and analysis period."""
-        return cls.from_data_and_datetimes(data, analysis_period.datetimes, header)
+        return cls.from_data_and_datetimes(
+            data, analysis_period.datetimes, header, middle_hour)
 
     @property
     def header(self):
@@ -94,6 +98,25 @@ class DataCollection(object):
     @header.setter
     def header(self, h):
         self._header = None if not h else Header.from_header(h)
+
+    @property
+    def middle_hour(self):
+        """Get or set whether the values are interpreted as falling on the half hour.
+
+        This is a boolean that sets whether the values in the data collection
+        are interpreted as falling on the half hour (True) or the hour (False).
+        The default is set to False to be interpreted as fallin on the hour.
+        Note that all EnergyPlus and Radiance results are totalled or averaged
+        values over the hour and so are better characterized by having this
+        input set to True."""
+        return self._middle_hour
+
+    @middle_hour.setter
+    def middle_hour(self, mh):
+        if mh:
+            assert isinstance(mh, bool), \
+                'Expected Boolean got {}'.format(type(mh))
+        self._middle_hour = False if not mh else mh
 
     def append(self, d):
         """Append a single item to the list."""
@@ -116,6 +139,19 @@ class DataCollection(object):
             'Expected Integer got {}'.format(type(i))
         self._data.insert(i, d)
 
+    def pop(self, i=-1):
+        """Remove the item at the given position in the data collection, and return it.
+
+        If no index is specified, a.pop() removes and returns the last
+        item in the list.
+        """
+        assert isinstance(i, int), \
+            'Expected Integer got {}'.format(type(i))
+        assert i < len(self._data), \
+            'Item({}) is larger than the length of the data collection({})' \
+            .format(i, len(self._data))
+        return self._data.pop(i)
+
     @property
     def datetimes(self):
         """Return datetimes for this collection as a tuple."""
@@ -123,14 +159,7 @@ class DataCollection(object):
 
     @property
     def values(self):
-        """Return the list of values.
-
-        Args:
-            header: A boolean that indicates if values should include the header
-
-        Return:
-            A list of values
-        """
+        """Return the list of values."""
         return self._data
 
     def duplicate(self):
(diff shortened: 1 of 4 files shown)


AntoineDao:
Tests were failing because tox.ini had an unexpected indent. Fixed and committed here. Let me know if I've done anything wrong.

PS: tests seem to say you've not broken anything @chriswmackey so that's pretty reassuring :)

ladybug/datacollection.py
```diff
@@ -327,22 +356,23 @@ def update_data_for_analysis_period(self, values, analysis_period):
         """
         return self.update_data_for_hours_of_year(values, analysis_period.hoys)
 
-    def interpolate_data(self, timestep, half_hour=False):
-        """Interpolate data for a finer timestep.
+    def interpolate_data(self, timestep, cumulative=False):
+        """Interpolate data for a finer timestep using a linear interpolation.
```
mostaphaRoudsari: Would be great to add a test for this. That will also make it clear how the cumulative Boolean works.

ladybug/epw.py
```diff
@@ -205,19 +205,21 @@ def _import_data(self, import_location_only=False):
...
+                        self._data[field_number].append(
+                            DataPoint(value, timestamp_hour))
 
                 line = epwin.readline()
 
+            # move last item to start position for fields on the hour
+            for field_number in xrange(self._num_of_fields):
+                middle_hour = EPWFields.field_by_number(field_number).middle_hour
+                if middle_hour is False:
+                    self._data[field_number].insert(0, self._data[field_number].pop())
```
mostaphaRoudsari: I suggest to write this in two separate lines so it's more clear what's happening.
```python
last_item = self._data[field_number].pop()
self._data[field_number].insert(0, last_item)
```

ladybug/wea.py
```diff
@@ -78,16 +78,23 @@ def from_json(cls, data):
...
+                Default is 1 for one value per hour.
         """
         epw = EPW(epwfile)
-        return cls(epw.location,
-                   epw.direct_normal_radiation,
-                   epw.diffuse_horizontal_radiation)
+        direct_normal = epw.direct_normal_radiation
+        diffuse_horizontal = epw.diffuse_horizontal_radiation
+        if timestep is not 1:
+            direct_normal = direct_normal.interpolate_data(timestep, True)
```
mostaphaRoudsari: I think we should add a warning here to let the user now current interpolation is linear and good for energy modeling however is not appropriate for daylight studies.

ladybug/epw.py
```diff
@@ -263,19 +265,16 @@ def save(self, folder=None, file_name=None):
             modEpwFile.writelines(self._header)
             lines = []
             try:
+                # move first item to end position for fields on the hour
+                for field in range(0, self._num_of_fields):
+                    middle_hour = EPWFields.field_by_number(field).middle_hour
+                    if middle_hour is False:
+                        self._data[field].append(self._data[field].pop(0))
```
mostaphaRoudsari: Shouldn't we put the data back in order after this change? If I understand what's going on here correctly if one writes the file a couple of time each time the list will be shifted for one item.

Also you want to have the original epw object after writing the modified epw file.

ladybug/epw.py
```diff
@@ -185,18 +186,17 @@ def _import_data(self, import_location_only=False):
                 # since I'm using this timestamp as the key and will be using it for
                 # sorting. I'm setting it up to 2015 - the real year will be collected
                 # under modelYear
-                timestamp_half_hour = DateTime(month, day, hour - 1)
-                timestamp_hour = DateTime(month, day, hour - 1)
+                timestamp_middle_hour = DateTime(month, day, hour - 1)
 
-                insert_hour = False
                 if month == 12 and day == 31 and hour == 24:
```
mostaphaRoudsari: What if instead of checking this here on the initial input we just adjust the hours all together at the end and only for data which doesn't follow middle hour convention? You already have to do that check to shift the last hour. Something like this should do it:

```py
# collect hourly data
for line in epwin:
    data = line.strip().split(',')
    year, month, day, hour = map(int, data[:4])

    # in an epw file year can be different for each month
    # since I'm using this timestamp as the key and will be using it for
    # sorting. I'm setting it up to 2015 - the real year will be collected
    # under modelYear
    timestamp = DateTime(month, day, hour - 1)

    for field_number in xrange(self._num_of_fields):

        value_type = EPWFields.field_by_number(field_number).value_type

        try:
            value = value_type(data[field_number])
        except ValueError as e:
            # failed to convert the value for the specific TypeError
            if value_type != int:
                raise ValueError(e)
            value = int(round(float(data[field_number])))

        self._data[field_number].append(DataPoint(value, timestamp_hour))

# move last item to start position for fields on the hour
for field_number in xrange(self._num_of_fields):
    middle_hour = EPWFields.field_by_number(field_number).middle_hour
    if middle_hour is False:
        # shift datetimes for an hour
        for data in self._data[field_number]:
            try:
                data.datetime = data.datetime.add_hour(1)
            except ValueError:
                # this is the last hour
                data.datetime = DateTime(1, 1, 0)

        # now move the last hour to first
        last_hour = self._data[field_number].pop()
        self._data[field_number].insert(0, last_hour)
```

chriswmackey:
Thanks @AntoineDao and @mostaphaRoudsari ,
All of these are great comments and fixes.  I agree with adding the half hour property to the header and I will get to work addressing the rest of the comments.
Thanks again!

ladybug/datacollection.py
```diff
@@ -14,11 +14,12 @@
 class DataCollection(object):
     """A list of data with a header."""
 
-    __slots__ = ('_header', '_data')
+    __slots__ = ('_header', '_data', '_middle_hour')
 
-    def __init__(self, data=None, header=None):
+    def __init__(self, data=None, header=None, middle_hour=None):
```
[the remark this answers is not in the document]
chriswmackey: I wholeheartedly agree with putting it on the header.  And it's ok if you didn't follow the comments about comparing data if you agree with the implementation.

ladybug/datacollection.py
```diff
@@ -95,6 +99,25 @@ def header(self):
...
+        This is a boolean that sets whether the values in the data collection
+        are interpreted as falling on the half hour (True) or the hour (False).
+        The default is set to False to be interpreted as fallin on the hour.
+        Note that all EnergyPlus and Radiance results are totalled or averaged
+        values over the hour and so are better characterized by having this
+        input set to True."""
+        return self._middle_hour
+
+    @middle_hour.setter
+    def middle_hour(self, mh):
```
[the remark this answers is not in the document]
chriswmackey: I originally thought it would be important but now that it's just a normal attribute on the header (like the other header attributes), it's definitely unnecessary.

ladybug/datacollection.py
chriswmackey: I might need some help on this one.  I'll give it a shot and post back here.

ladybug/epw.py
chriswmackey: ok

ladybug/wea.py
chriswmackey: Integrated a warning into the docsting.  Let me know if it belongs anywhere else.

ladybug/epw.py
chriswmackey: Good catch and that's a bug uncaught by our tests.  Let me know if you think that editing and fixing the epw object before and afterwards is the best approach (as opposed to trying to changing the data written into the file as it is being iterated through).  Seeing what you did below, I would guess that you would say editing it and then putting it back is conceptually clearer.

ladybug/epw.py
chriswmackey: Ok.  I can see how this is just as efficient as the implementation I had before but it's easier to read.  I've changed it to be just like this.

ladybug/datacollection.py
```diff
@@ -392,7 +369,7 @@ def interpolate_data(self, timestep, cumulative=False):
                 _data[i].value = d.value / timestep
 
         # shift data if half-hour interpolation has been selected.
-        if self.middle_hour is True:
+        if self.header.middle_hour is True:
```
mostaphaRoudsari: You need a check here to ensure header is not None before checking for middle_hour unless we are generating an empty header if not provided.

ladybug/datacollection.py
chriswmackey: Now that I am going through this, I think it might be better if I submit a test as a separate pull request after we merge this one in.  Given that this is the first test I'd be submitting, I would like you and @AntoineDao to get a clear look at it.

ladybug/wea.py
mostaphaRoudsari: Looks good! This will be great for developers who will see the source code but I think we should also make it explicit and print out the warning for users if timestep is not 1 and the wea is created from an epw file.

ladybug/datacollection.py
mostaphaRoudsari: Sure! let me know.

ladybug/wea.py
chriswmackey: Understood.  Yes, when I expose it on the component, I will add a warning.

chriswmackey: ... or do you mean that we should be printing a warning?

ladybug/datacollection.py
chriswmackey: Ok.  I committed a test.  It worked when I ran it locally.  Hopefully it makes sense.

PR #56, opened 2018-07-09
Increasing coverage!!!

chriswmackey:
Added some more Wea tests.

I also added a new property to the Wea object that I think will be useful (it's at least very useful for the tests).

@mostaphaRoudsari , let me know if you see anything glaringly wrong in the tests.

### DIFF
--- a/tests/wea_test.py
+++ b/tests/wea_test.py
@@ -1,7 +1,8 @@
 # coding=utf-8
 
 import unittest
-import os
+import pytest
+
 from ladybug.wea import Wea
 
 
@@ -57,6 +58,7 @@ class WeaTestCase(unittest.TestCase):
 
         assert wea_from_zh.location.city == 'Chicago Ohare Intl Ap'
         assert wea_from_zh.timestep == 1
+        # TODO: Add checks for the values produced by the model
 
     def test_json_methods(self):
         """Test JSON serialization methods"""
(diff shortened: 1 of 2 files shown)


tests/wea_test.py
```diff
@@ -33,6 +34,32 @@ def test_from_stat(self):
...
 
+    def test_from_clear_sky(self):
+        """Test from original clear sky"""
+        stat_path = './tests/stat/chicago.stat'
+        wea_from_stat = Wea.from_stat_file(stat_path)
+        location = wea_from_stat.location
+        wea_from_clear_sky = Wea.from_ashrae_clear_sky(location)
+
+        assert wea_from_clear_sky.location.city == 'Chicago Ohare Intl Ap'
+        assert wea_from_clear_sky.timestep == 1
```
mostaphaRoudsari: We most likely want to check a number of values that are calculated as well as `city` and `timestep`.

tests/wea_test.py
```diff
@@ -33,6 +34,32 @@ def test_from_stat(self):
...
+
+        cc = [0.5] * 8760
+        rh = [50] * 8760
+        dbt = [20] * 8760
+        ws = [2] * 8760
+        wea_from_zh = Wea.from_zhang_huang_solar_model(location, cc, rh, dbt, ws)
+
+        assert wea_from_zh.location.city == 'Chicago Ohare Intl Ap'
+        assert wea_from_zh.timestep == 1
+        # TODO: Add checks for the values produced by the model
```
mostaphaRoudsari: Agreed!

tests/wea_test.py
```diff
@@ -65,8 +92,37 @@ def test_import_stat(self):
...
+        wea_from_stat = Wea.from_stat_file(stat_path)
+
+        diffuse_horiz_rad = wea_from_stat.diffuse_horizontal_radiation
+        direct_horiz_rad = wea_from_stat.direct_horizontal_radiation
+        glob_horiz_rad = wea_from_stat.global_horizontal_radiation
+
+        assert [x.value for x in glob_horiz_rad] == pytest.approx(
+            [x + y for x, y in zip(diffuse_horiz_rad, direct_horiz_rad)], rel=1e-3)
+
     def test_radiation_on_surface(self):
```
mostaphaRoudsari: This comment is unrelated to test PR.

I think we don't want to use term `surface` under the wea class. It can be confusing to have radiation calculation on a surface under Wea class. Think about someone searching the API for keyword radiation and finds this. It can be misleading. I would suggest to rename this method to `radiation_at_angle` or something like that.

PR #57, opened 2018-07-09
Finished adding support for leap year

mostaphaRoudsari:
- Added support for leap year to Datetime
- Added support for leap year to sunpath
- Fixed spelling mistake in method name: calculate_sun_from_date_time

### DIFF
--- a/ladybug/dt.py
+++ b/ladybug/dt.py
@@ -70,7 +70,7 @@ class DateTime(datetime):
         Args:
             hoy: A float value 0 <= and < 8760
         """
-        return cls.from_moy(round(hoy * 60))
+        return cls.from_moy(round(hoy * 60), leap_year)
 
     @classmethod
     def from_moy(cls, moy, leap_year=False):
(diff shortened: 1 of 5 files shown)


mostaphaRoudsari:
@chriswmackey, I will push a fix soon!

ladybug/dt.py
```diff
@@ -79,10 +79,15 @@ def from_moy(cls, moy, leap_year=False):
...
         """
-        num_of_minutes_until_month = (0, 44640, 84960, 129600, 172800, 217440,
-                                      260640, 305280, 349920, 393120, 437760,
-                                      480960, 525600)
-
+        if not leap_year:
+            num_of_minutes_until_month = (0, 44640, 84960, 129600, 172800, 217440,
+                                          260640, 305280, 349920, 393120, 437760,
+                                          480960, 525600)
+        else:
```
chriswmackey: I was thinking that this would look cleaner by just declaring 1440 as a variable once but then I realized that would unnecessarily create an extra variable that gets kept in memory.  So this is good.

PR #58, opened 2018-07-10
Improved Wea Tests

chriswmackey:
@mostaphaRoudsari ,

I hope that you don't mind that I just merge this one in myself given that these are all things we talked about yesterday.

### DIFF
--- a/ladybug/wea.py
+++ b/ladybug/wea.py
@@ -496,20 +496,19 @@ class Wea(object):
         hoy = int(hoy * self.timestep)
         return self.direct_normal_radiation[hoy], self.diffuse_horizontal_radiation[hoy]
 
-    def radiation_on_surface(self, surface_altitude=90, surface_azimuth=180,
-                             ground_reflectance=0.2, isotrophic=True):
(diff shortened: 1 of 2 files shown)


chriswmackey:
@mostaphaRoudsari ,
It seems that I need your review anyway.  Also, I am not sure how coverage decreased when all I did was improve some of the existing tests.  Maybe there is a stray blank line in there?  It would be nice to have a threshold below which a loss of coverage is not counted as a loss.

ladybug/wea.py
```diff
@@ -496,20 +496,19 @@ def get_radiation_values_for_hoy(self, hoy):
         hoy = int(hoy * self.timestep)
         return self.direct_normal_radiation[hoy], self.diffuse_horizontal_radiation[hoy]
 
-    def radiation_on_surface(self, surface_altitude=90, surface_azimuth=180,
-                             ground_reflectance=0.2, isotrophic=True):
-        """Returns the total radiation falling on a surface at each timestep.
+    def directional_radiation(self, altitude=90, azimuth=180,
```
mostaphaRoudsari: 👍

PR #60, opened 2018-07-12
Switched to python2.7-alpine base image

AntoineDao:
Noticed docker images were built on standard python2.7 image. If the alpine (tiny container os) version is used the size of the final container can be dramatically reduced (which is good when scaling containerised applications). Point in case below:
image
It will also drastically reduce the honeybee docker image by default. I gave it a quick test and it can still install radiance no problem (will need to check if it can still run radiance properly but I see no reason why it couldn't).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7
+FROM python:2.7-alpine
 MAINTAINER Ladybug Tools "info@ladybug.tools"
 
 WORKDIR /usr/local/lib/python2.7/site-packages


AntoineDao:
I have no idea how I increased coverage on this!

mostaphaRoudsari:
Let me copy it here:

> Radiance [NREL distribution] won't work on Alpine for similar reason as this: https://github.com/moby/moby/issues/14972#issuecomment-124784587
> I wonder if it is possible to build radiance against `uclibc`.

PR #61, opened 2018-07-12
Finished supporting leap year in Wea

mostaphaRoudsari:
This commit finishes the development of Wea class to suppoer leap year. It was mainly neccassary to create clear skies for studies. I also cleaned the code and removed some duplicated functionalities in a single method. Other changes are:

1. Wea now accepts DataCollections and not lists. I added a classmethod to support creating Wea from list of numbers.
2. For timestep 1 the datetime is adjusted for 30 minutes for all the classmethods. The __init__ doesn't modify datetimes. This makes the class cleaner and better structured. Also the write method is now clean and doesn't have multiple checks.

### DIFF
--- a/ladybug/wea.py
+++ b/ladybug/wea.py
@@ -24,25 +24,44 @@ class Wea(object):
 
     Attributes:
         location: Ladybug location object.
-        direct_normal_radiation: A list of direct normal radiation values for
-            every hourly timestep of the year.
-        diffuse_horizontal_radiation: A list of diffuse horizontal radiation
-            values for every hourly timestep of the year.
+        direct_normal_radiation: An annual DataCollection of direct normal radiation
+            values.
+        diffuse_horizontal_radiation: An annual DataCollection of diffuse horizontal
+            radiation values for every hourly timestep of the year.
         timestep: An optional integer to set the number of time steps per hour.
             Default is 1 for one value per hour.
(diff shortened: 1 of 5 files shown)


ladybug/wea.py
```diff
@@ -246,48 +279,51 @@ def from_ashrae_clear_sky(cls, location, sky_clearness=1, timestep=1):
...
             months.append(sun.datetime.month - 1)
             altitudes.append(sun.altitude)
 
         # compute hourly direct normal and diffuse horizontal radiation
         direct_norm_rad, diffuse_horiz_rad = \
-            cls._get_data_collections(location, timestep)
+            cls._get_empty_data_collections(location, timestep, is_leap_year)
 
         for i, alt in enumerate(altitudes):
-            if alt > 0.1:
```
chriswmackey: I think taking this out might cause an error since it was causing an exception to be thrown after it computes e raised to an extremely high exponent that cannot be expressed in 16-bit float numbers.  If it's passing the checks, though, maybe I am just being paranoid.  Let's merge and I'll do some checks of my own.

mostaphaRoudsari: This should handle the exception:
https://github.com/ladybug-tools/ladybug/pull/61/files#diff-f268b272e07e1ae230a4c18677efa6cbR307

PR #62, opened 2018-07-13
Put Back Some Code in Read Stat

chriswmackey:
@mostapharoudsari , it seems that one of the changes you made a few days back to upgrade to Python 3 broke the ability to read the degree symbol that is in a lot of stat files.  I'm putting back the code to its original stat with this pull request but let me know if  you find a better way to do it.

### DIFF
--- a/ladybug/stat.py
+++ b/ladybug/stat.py
@@ -63,7 +63,7 @@ class Stat(object):
     def import_data(self):
         """Import data from a stat file.
         """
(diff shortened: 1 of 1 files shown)


mostaphaRoudsari:
Hi @chriswmackey, That was intentional to work both on Python 2 and Python 3. As you can see now that you put it back it fails against Python 3. Can you share a sample file that doesn't work so I can give it a try? We should also add that file to test.

PR #63, opened 2018-07-13
Added a check for IronPython

mostaphaRoudsari:
IronPython doesn't handle codecs.open errors correctly. This commits adds an extra check to fix this issue.  
@chriswmackey this should resolve #62

### DIFF
--- a/ladybug/stat.py
+++ b/ladybug/stat.py
@@ -3,6 +3,7 @@ from .location import Location
 import os
 import re
 import codecs
+import platform
 
 
 class Stat(object):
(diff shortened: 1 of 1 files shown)


mostaphaRoudsari:
DO NOT MERGE! It fails in older versions of IronPython that comes with Rhino 5! I will send a fix soon.

chriswmackey:
Thanks for addressing this @mostaphaRoudsari .  I'm looking forward to the next commit/

mostaphaRoudsari:
This one should do it for Python 2, 3 and IronPython new and old! Oh well.

PR #68, opened 2018-08-13
Set Default IronPython Variable for Very Old Versions

chriswmackey:
The old version of IronPython that ships with Rhino 5 has some bugs in the platform module.  In this case, the inron_python variable does not get set at all.  To avoid this case, I set a default variable of iron_python to True before checking the platform module.

### DIFF
--- a/ladybug/stat.py
+++ b/ladybug/stat.py
@@ -63,7 +63,7 @@ class Stat(object):
     def import_data(self):
         """Import data from a stat file.
         """
-        statwin = codecs.open(self.file_path, 'r', encoding='utf-8', errors='ignore')
+        statwin = codecs.open(self.file_path, 'r')
         try:
             line = statwin.readline()
             # import header with location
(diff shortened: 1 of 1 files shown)


ladybug/stat.py
```diff
@@ -64,6 +64,8 @@ def is_data_loaded(self):
     def import_data(self):
         """Import data from a stat file.
         """
+        # set default state to ironpython for very old ironpython (2.7.0)
+        iron_python = True
```
mostaphaRoudsari: Hey @chriswmackey, did you have a case that fails? This is what the other part of the code is supposed to handle:

https://github.com/ladybug-tools/ladybug/blob/9a6cb08dd122e1c3a7782904afb45864596bbf97/ladybug/stat.py#L72-L77

mostaphaRoudsari: Here is the answer to my question:
https://discourse.ladybug.tools/t/error-tauclearsky-component/3495/9

PR #69, opened 2018-08-23
Added get_highest_values to datacollection

TobySkov:
https://github.com/ladybug-tools/honeybee/issues/237


TobySkov:
Hi @mostaphaRoudsari, where can I find your comments ? 

Have added a test to datacollection_test.py

tests/datacollection_test.py
```diff
@@ -60,7 +60,17 @@ def test_json_methods(self):
...
-
+        
+    def test_get_highest_values(self):
+    # To test get_highest_values, a range of yearly-hour values will be used
+        test_data = range(8760)
+        test_number_of_highest_values = len(test_data)/2
+        
+        test_highest_values, test_is_data_in_highest_values = DataCollection.get_highest_values(test_data, test_number_of_highest_values)
+        
+        assert test_highest_values == list(reversed(range(4380,8760)))
```
mostaphaRoudsari: range(8759, 4381, -1)?

tests/datacollection_test.py
```diff
@@ -60,7 +60,17 @@ def test_json_methods(self):
...
+        assert test_is_data_in_highest_values == [0]*4380+[1]*4380
```
mostaphaRoudsari: spaces are missing.

ladybug/datacollection.py
```diff
@@ -163,13 +163,13 @@ def get_highest_values(data, number_of_highest_values):
...
-        highest_values_index = sorted(range(lenght_data), key = lambda k: data[k], reverse = True)[0:number_of_highest_values]
+        highest_values_index = sorted(range(lenght_data), key = lambda k: data[k], reverse = True)[0:int(number_of_highest_values)]
         
         is_data_in_highest_values = [0]*lenght_data 
-        for i in range(number_of_highest_values):
+        for i in range(int(number_of_highest_values)):
             is_data_in_highest_values[highest_values_index[i]]=1
             
-        highest_values = sorted(data, reverse = True)[0:number_of_highest_values]
+        highest_values = sorted(data, reverse = True)[0:int(number_of_highest_values)]
```
mostaphaRoudsari: What if you add `number_of_highest_values = int(number_of_highest_values)` once on the top and then use the integer value?

mostaphaRoudsari:
@TobySkov thanks. I forgot to submit the review. Now you should be able to see them. Cheers.

ladybug/datacollection.py
```diff
@@ -148,6 +148,30 @@ def average(data):
         """Return average value for a list of ladybug data."""
         values = (value.value for value in data)
         return sum(values) / len(data)
+    
+    @staticmethod
```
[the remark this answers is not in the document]
TobySkov: Drawing inspiration from other methods in the class, isn't self unnecessary in this case? Or would you prefer to have the output as attributes for the instance?

ladybug/datacollection.py
```diff
@@ -148,6 +148,30 @@ def average(data):
...
+    def get_highest_values(data, number_of_highest_values):
+        """
+        Args:
+            data: A list of DataPoint to be processed
+            number_of_highest_values: Number of highest values to account for
+            
+        Return:
+            highest_values: The n highest values in data list, ordered from
+                highest to lowest
+            is_data_in_highest_values: Binary list with same lenght as data
```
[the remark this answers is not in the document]
TobySkov: Would the new method also go in datacollection.py?

TobySkov:
Thank you for the comments @mostaphaRoudsari 
I have made some questions for your comments - let me know what you think

ladybug/datacollection.py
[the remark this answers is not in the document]
mostaphaRoudsari: I think it fits better to be a normal method. We should probably make most of other staticmethods in this class as just normal methods. I don't see many cases that one wants to use this method without creating an instance.

ladybug/datacollection.py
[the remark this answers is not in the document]
mostaphaRoudsari: no. that will be part of where you need that functionality. Since that is not a common usecase I suggest to keep it outside.

TobySkov:
All initial comments are implemented :)

ladybug/datacollection.py
```diff
@@ -163,16 +162,27 @@ def get_highest_values(data, count):
             highest_values_index: Indicies of the n highest values in data 
                 list, ordered from highest to lowest
         """
-        lenght_data = len(data)
+        dataPoints = self._data
+        
+        values = list(obj._value for obj in dataPoints)
```
mostaphaRoudsari: `values = [obj._value for obj in dataPoints]` is better way of writing this.

ladybug/datacollection.py
```diff
@@ -163,16 +162,27 @@ def get_highest_values(data, count):
...
         count = int(count)
         
-        highest_values = sorted(data, reverse = True)[0:count]
+        assert count <= lenght_values, \
+            'Count must be equal to or smaller than list of data lenght'
+            
+        assert count > 0, \
+            'Count must be higher than zero'
+        
+        self.highest_values = sorted(values, reverse = True)[0:count]
```
mostaphaRoudsari: I don't think you need to assign `highest_values` to class as a property. This is a once tot be used data and also can change based on input for `count`.

ladybug/datacollection.py
```diff
@@ -163,16 +162,27 @@ def get_highest_values(data, count):
...
         
-        highest_values_index = sorted(range(lenght_data), key = lambda k: data[k],
-                                                         reverse = True)[0:count]
+        self.highest_values_index = sorted(range(lenght_values),
```
mostaphaRoudsari: same comment here!

ladybug/datacollection.py
```diff
@@ -14,7 +14,7 @@
 class DataCollection(object):
     """A list of data with a header."""
 
-    __slots__ = ('_header', '_data')
+    __slots__ = ('_header', '_data', 'highest_values', 'highest_values_index')
```
mostaphaRoudsari: I commented down there. Let's not add new properties to class as they are not really neccassary for all the instances of the class.

ladybug/datacollection.py
mostaphaRoudsari: Also maybe we should add a property for this that only returns the values without dates and another one that only return dates. I'm commenting on these during the workshop and didn't get the chance to see if they are already there.

TobySkov:
Thank you again for a very nice workshop in Denmark ! Can't assess from your comments if you want to return dates to complement the indicies or not?

mostaphaRoudsari:
@TobySkov, Look good! Thank you for the changes and thank you for being patient. I'm going to merge this in.

> Can't assess from your comments if you want to return dates to complement the indicies or not?

That could have been an option but one can find the dates using the index list. Another option would be to return the DataPoint instead of DataPoint.value which already has datetime data.

PR #70, opened 2018-09-04
Added Design Day Parsing Capabilities

chriswmackey:
@mostaphaRoudsari ,
This is that rather large pull request that I had mentioned to you.  It adds the following capabilities:

- Ability to read in .ddy files and produce editable Design Day objects.
- Create Design Day objects from raw parameters
- Write edited design day objects to a new .ddy file
- Get annual design day objects from the .stat file (essentially allowing us to import most of the other useful data in the stat file that we do not currently import in Ladybug Legacy).
- Produce monthly Design Day objects from stat files, which engineers are what engineers are often using to size zone equipment
- Get hourly data collections of conditions over the design day (this will eventually allow users to visualize design conditions in the same way that they visualize the hourly EPW data).  This may also eventually support the creation of EPW files composed entirely of design days, which can be useful for studies like passive survivability studies.

In addition to this, I tired to increase coverage by adding checks for all new code and filling in checks for the psychrometric functions.


chriswmackey:
@mostaphaRoudsari ,
I just need y ou help in understanding how the config module should be imported for python 3, it seems.

mostaphaRoudsari:
@chriswmackey is this PR ready to be reviewed?

chriswmackey:
Now it's ready for review, @mostaphaRoudsari !

chriswmackey:
@mostaphaRoudsari ,
The data and values attributes have been switched on the datacollection so it's all ready for your review now.

ladybug/psychrometrics.py
```diff
@@ -29,76 +28,58 @@ def find_saturated_vapor_pressure_high_accuracy(t_kelvin):
...
-    return enthalpy
+    enthalpy = ((1.01 + (1.89 * humid_ratio)) * db_temp) + (2500 * humid_ratio)
+    return enthalpy if enthalpy >= 0 else 0
 
 
-def find_wet_bulb(db_temp, rh, psta=101325):
-    """
-    Calculates Wet Bulb Temperature (C) at Temperature db_temp (C),
-    Relative Humidity rh (%), and Barometric Pressure psta (Pa).
+def wet_bulb_from_db_rh(db_temp, rh, b_press=101325):
```
mostaphaRoudsari: I'm in favor of name changes and removing *find* or *calculate* form method names but do we really need *from_x_y* after the names? If there are 5 different methods for wetbulb calculation then we probably want to differentiate them by add *from_x_y* similar to how dew_point methods are but in this case and other cases like enthalpy I'm not sure if we need those.

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+# atmospheric extinction coefficient
+monthly_b = [0.141, 0.142, 0.149, 0.164, 0.177, 0.185, 0.186, 0.182,
+             0.165, 0.152, 0.144, 0.141]
+
+
+def ashrae_clear_sky(altitudes, month, sky_clearness=1):
+    """Calculate solar flux for an original ASHRAE Clear Sky
+
+    Args:
+        altitudes = A list of solar altitudes in degrees
```
mostaphaRoudsari: Should be `altitudes:` not `altitudes =`

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
+"""Functions for computing radiation for different idealized skies"""
+import math
+
+
+"""ORIGINAL AHSRAE CLEAR SKY SOLAR MODEL"""
+# apparent solar irradiation at air mass m = 0
+monthly_a = [1202, 1187, 1164, 1130, 1106, 1092, 1093, 1107, 1136,
```
mostaphaRoudsari: Since these are Constants a good practice is to have them all caps: MONTHLYA. Also since you don't want someone to change them I would change them to be a tuple and not a list. Same comments about MONTHLYB.

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+            of the connected altitudes in W/m2.
+    """
+    dir_norm_rad = []
+    dif_horiz_rad = []
+
+    ab = 1.219 - (0.043 * tb) - (0.151 * td) - (0.204 * tb * td)
+    ad = 0.202 + (0.852 * tb) - (0.007 * td) - (0.357 * tb * td)
+    for alt in altitudes:
+        # calculate hourly air mass between top of the atmosphere and earth
+        air_mass = 0
```
mostaphaRoudsari: Do we need to set air_mass to 0 here?

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+            dif_horiz_rad.append(0)
+
+    return dir_norm_rad, dif_horiz_rad
+
+
+"""ZHANG-HUANG SOLAR MODEL"""
+# TODO: Split golbal into direct and diffuse using Perez method.
+
+# extraterrestrial solar constant (W/m2)
+irr0 = 1355
```
mostaphaRoudsari: 1. Let's use ALL CAPS for constant values.
2. If these values are only used by one method I would move them under the method to ensure it will be: 1. protected from being removed by accident or being overwritten as a global value. 2. copy/pasted with the method.

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+        alt = A solar altitudes in degrees.
+        cloud_cover: A float value between 0 and 10 that represents the sky cloud cover
+            in tenths (0 = clear; 10 = completely overcast)
+        relative_humidity: A float value between 0 and 100 that represents
+            the relative humidity in percent.
+        dry_bulb_present: A float value that represents the dry bulb
+            temperature at the time of interest (in degrees C).
+        dry_bulb_t3_hrs: A float value that represents the dry bulb
+            temperature at three hours before the time of interest (in degrees C).
+        wind_speed: A float value that represents the wind speed in m\s.
```
mostaphaRoudsari: m/s ?

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+                k_ds = (3.996 - (3.862 * sin_alt) +
+                        (1.540 * sin_alt**2)) * k_t**3
+            diff_ir = (irr0 * sin_alt * (k_t - k_ds)) / (1 - k_ds)
+            dir_horiz_ir = (irr0 * sin_alt * k_ds * (1 - k_t)) / (1 - k_ds)
+            dir_ir = dir_horiz_ir / math.sin(math.radians(alt))
+
+    return dir_ir, diff_ir
+
+
+"""ENERGYPLUS HORIZONTAL INFRARED INTESDITY MODEL"""
```
mostaphaRoudsari: D should be removed from INTESDITY?

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+# stefan-boltzmann constant
+sigma = 5.6697e-8
```
mostaphaRoudsari: Same comment about having constant values as global variables.

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+def energyplus_horizontal_infrared(sky_cover, dry_bulb, dew_point):
```
mostaphaRoudsari: do we need energyplus in the name?
If we do can we move it to the end of the name so if someone looks for horizontal the autocomplete suggests this function?

ladybug/skymodel.py
```diff
@@ -0,0 +1,205 @@
...
+    db_k = dry_bulb + 273.15
+    dp_k = dew_point + 273.15
+    sky_emiss = (0.787 + (0.764 * math.log(dp_k / 273.15))) * \
+        (1 + (0.022 * sky_cover) - (0.0035 * (sky_cover ** 2)) +
+         (0.00028 * (sky_cover ** 3)))
+    horiz_ir = sky_emiss * sigma * (db_k ** 4)
+    return horiz_ir
+
+
+def energyplus_sky_temperature(horiz_ir, dry_bulb):
```
mostaphaRoudsari: Same comment about having EnergyPlus in the name.

tests/designday_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+    def test_standard_ddy_properties(self):
+        """Test properties of a standard ddy."""
+        relative_path = './tests/ddy/tokyo.ddy'
+
+        ddy = Ddy.from_ddy_file(relative_path)
+
+        # Test accuracy of import
+        assert ddy.location.city == 'TOKYO HYAKURI_JPN Design_Conditions'
+        assert -0.1 < ddy.location.latitude - 36.18 < 0.1
```
mostaphaRoudsari: You might want consider `pytets.approx` in such cases: https://docs.pytest.org/en/latest/reference.html#pytest-approx
`assert ddy.location.latitude - 36.18 = pytets.approx(0.1)`

tests/psychrometrics_test.py
```diff
@@ -1,12 +1,21 @@
 # coding=utf-8
 
 import unittest
-from ladybug.psychrometrics import find_saturated_vapor_pressure_torr, \
-    find_enthalpy, \
-    find_wet_bulb, \
-    find_dew_point, \
-    find_rel_humid_from_dry_bulb_dew_pt, \
-    find_air_temp_from_wet_bulb
+from ladybug.psychrometrics import \
```
mostaphaRoudsari: I feel `import ladybug.psychrometrics as psy` and then using `psy.function_name` will be cleaner but this will also work.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+from .psychrometrics import dew_point_from_db_enth
+from .psychrometrics import dew_point_from_db_wb
+
+import math
+import os
+import re
+import codecs
+import platform
+
+day_types = ['SummerDesignDay', 'WinterDesignDay', 'Sunday', 'Monday',
```
mostaphaRoudsari: Should we move all this inside the Class as a class CONSTANT? Or you can put them under a property.
```Python
@property
def day_types:
    return ('SummerDesignDay', 'WinterDesignDay', 'Sunday', 'Monday', ...)
```
```

File: ladybug/designday.py [hunk 14]
```diff
@@ -0,0 +1,1109 @@
...
+        location: A Ladybug location object
+        design_days: A list of the design days in the ddy file.
+        file_path: Optional file path into which ddy file can be written.
+    """
+    # TODO: set the default file path to use the ladybug default folder.
+    def __init__(self, location, design_days=[], file_path=None):
+        """Initalize the class."""
+        self.location = location
+        self.design_days = design_days
+        self.file_path = file_path or 'C:\\ladybug\\unnamed.ddy'
```
mostaphaRoudsari: Why do you need to set file_path here?

In any case we should not set it up to a Windows format file path.
```self.file_path = file_path or 'unnamed.ddy'```

File: ladybug/designday.py [hunk 15]
```diff
@@ -0,0 +1,1109 @@
...
+
+    @classmethod
+    def from_ddy_file(cls, file_path):
+        """Initalize from a ddy file object from an existing ddy file.
+
+        args:
+            file_path: A full string to the .ddy file.
```
mostaphaRoudsari: full string?

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+        """
+        # defaults
+        location = None
+        design_days = []
+
+        # check that the file is there
+        if not os.path.isfile(file_path):
+            raise ValueError(
+                'Cannot find an ddy file at {}'.format(file_path))
```
mostaphaRoudsari: **a** ddy?

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+
+    @classmethod
+    def from_ddy_file(cls, file_path):
+        """Initalize from a ddy file object from an existing ddy file.
+
+        args:
+            file_path: A full string to the .ddy file.
+        """
+        # defaults
+        location = None
```
mostaphaRoudsari: Why would you need that here?

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+class Ddy(object):
+    """Import data from a local .ddy file.
+
+    properties:
+        location: A Ladybug location object
+        design_days: A list of the design days in the ddy file.
+        file_path: Optional file path into which ddy file can be written.
+    """
+    # TODO: set the default file path to use the ladybug default folder.
+    def __init__(self, location, design_days=[], file_path=None):
```
mostaphaRoudsari: Doesn't design_days already include location data? Based on `from_design_day` it does. Then the question will be should we remove location input and get it from design_days?

Also I'm not sure what does a DDY class with no design_days mean? Should we remove empty list as the default input?

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+    def from_design_day(cls, design_day):
+        """Initalize from a ddy file object from a ladybug design day object.
+        """
+        return cls(design_day.location, [design_day])
+
+    def save(self, folder=None, file_name=None):
+        """Save ddy object as a .ddy file."""
+        # check file_path
+        if not folder:
+            folder = self.folder
```
mostaphaRoudsari: If you're keeping self.folder for the purpose of saving the file then I think you can just remove it and make folder as a required input. The same about file_name.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+            self.location.city, str(len(self._design_days)))
+
+
+# TODO: add a property for hourly horizontal infrared radiation.
+class DesignDay(object):
+    """An object representing design day conditions.
+
+    properties:
+        name: A text string to set the name of the design day
+        day_type: Choose from 'SummerDesignDay', 'WinterDesignDay' or other
```
mostaphaRoudsari: What are other EnergyPlus days and how can one get it? Now that I see this I think it is good to have it as a property or something like that.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+
+    @classmethod
+    def from_ep_string(cls, ep_string, location):
+        """Initalize from an EnergyPlus string of a SizingPeriod:DesignDay.
+
+        args:
+            ep_string: A full string representing a SizingPeriod:DesignDay.
+        """
+        # format the object into a list of properties
+        ep_lines = ep_string.split('\n')
```
mostaphaRoudsari: I would use regular expression here. This method can fail.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+        # assign sky condition values based on the solar model
+        if self.sky_condition.solar_model == 'ASHRAEClearSky':
+            ep_vals[25] = self.sky_condition.clearness
+        if self.sky_condition.solar_model == 'ASHRAETau':
+            ep_vals[23] = self.sky_condition.tau_b
+            ep_vals[24] = self.sky_condition._tau_d
+            ep_vals.pop()
+
+        # put everything together into one string
+        ep_string = ' SizingPeriod:DesignDay,\n'
```
mostaphaRoudsari: Can we use string formatting to put this together? This is currently really hard to read.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+
+    @property
+    def location(self):
+        """Get or set the location."""
+        return self._location
+
+    @location.setter
+    def location(self, data):
+        assert hasattr(data, 'isLocation'), 'Expected' \
+            ' isLocation type. Got {}'.format(str(data))
```
mostaphaRoudsari: 1. `type(data)` will be more informative than `str(data)`.
2. format does convert the type to string automatically.
3. Expected Location not isLocation.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+            self.sky_condition.month,
+            self.sky_condition.day_of_month,
+            23)
+
+    @property
+    def hourly_datetimes(self):
+        return self._sky_condition._get_datetimes()
+
+    @property
+    def hourly_dry_bulb_data(self):
```
mostaphaRoudsari: Do we need data at the end of the name for all these methods?

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+        self.modifier_schedule = str(modifier_schedule)
+
+    @property
+    def hourly_values(self):
+        """A list of temperature values for each hour over the design day."""
+        return [self._max_dry_bulb - self._dry_bulb_range * x for
+                x in temp_multipliers]
+
+    @property
+    def max_dry_bulb(self):
```
mostaphaRoudsari: I would suggest `dry_bulb_max`.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+        wet_bulb_range: Optional wet bulb temperature range
+    """
+    def __init__(self, hum_type, hum_value, barometric_pressure=101325,
+                 schedule='', wet_bulb_range=''):
+        """Initalize the class."""
+        self.hum_type = hum_type
+        self.hum_value = hum_value
+        self.barometric_pressure = barometric_pressure
+        self.schedule = schedule
+        self.wet_bulb_range = wet_bulb_range
```
mostaphaRoudsari: Should it be a list and not a string?  I mean if it is a range.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+    """Represents wind and rain conditions on the design day.
+
+    attributes:
+        wind_speed
+        wind_direction
+        rain
+        snow_on_ground
+    """
+    def __init__(self, wind_speed, wind_direction=0,
+                 rain='No', snow_on_ground='No'):
```
mostaphaRoudsari: I would use True and False as default values and convert them to 'Yes', 'No' under the hood.

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+        """Initalize the class."""
+        self.wind_speed = wind_speed
+        self.wind_direction = wind_direction
+        self.rain = rain
+        self.snow_on_ground = snow_on_ground
+
+    @property
+    def hourly_values(self):
+        """A list of wind speed values for each hour over the design day."""
+        return [self._wind_speed for i in range(24)]
```
mostaphaRoudsari: what about [self._wind_speed] * 24?

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+
+    @property
+    def hourly_wind_dirs(self):
+        """A list of wind directions for each hour over the design day."""
+        return [self._wind_direction for i in range(24)]
```
mostaphaRoudsari: The same here.

ladybug/stat.py
```diff
@@ -130,20 +161,87 @@ def import_data(self):
...
+                    winter_keys = self._split_des_day_row(
+                        line, 'Design Stat	Coldest')
+                elif winter_des_section is True and 'Heating' in line:
+                    winter_des_section = False
+                    winter_vals = self._split_des_day_row(
+                        line, 'Heating')
+                    self._winter_des_day_dict = {}
+                    for key, val in zip(winter_keys, winter_vals):
+                        self._winter_des_day_dict[key] = val
+                elif 'Design Stat	Hottest' in line:
```
mostaphaRoudsari: What about using regular expression to avoid all these if, elif statements?

ladybug/stat.py
```diff
@@ -20,11 +32,35 @@ class Stat(object):
...
+        self._monthly_wb_100 = []
+        self._monthly_db_range_50 = []
+        self._monthly_wb_range_50 = []
+        self._monthly_wind = []
+        self._monthly_wind_dirs = []
+        self._monthly_tau_beam = []
+        self._monthly_tau_diffuse = []
+
+        # import the data from the file
+        self.import_data()
```
mostaphaRoudsari: You most likely want to make `import_data` internal e.g. `_import_data`. If someone calls this method on an instance you will end up with duplicate input data.

ladybug/skymodel.py
chriswmackey: 1. I wasn't aware of the ALL CAPS convention with constants in PEP8.  Thanks for pointing that out.
2. I'll move it under the function but I wasn't sure if this creates any inefficiency in the memory required to store the constant value given that that function will likely get called many thousands of times.    If it does create inefficiency, I guess the little loss of memory is worth the more readable code.

mostaphaRoudsari: > but I wasn't sure if this creates any inefficiency in the memory required to store the constant value given that that function will likely get called many thousands of times.

Since it is a property of the class itself all the instances will point into the same place in memory. So it doesn't!

ladybug/skymodel.py
chriswmackey: I have removed it and just referenced energyplus in the description.

tests/designday_test.py
chriswmackey: Ok, I've updated all test like this to use pytest

ladybug/designday.py
```diff
@@ -0,0 +1,1109 @@
...
+            '!- Diffuse Solar Day Schedule Name',
+            '!- ASHRAE Clear Sky Optical Depth for Beam Irradiance (taub)',
+            '!- ASHRAE Clear Sky Optical Depth for Diffuse Irradiance (taud)',
+            '!- Clearness [0.0 to 1.2]']
+temp_multipliers = [0.82, 0.88, 0.92, 0.95, 0.98, 1, 0.98, 0.91, 0.74,
+                    0.55, 0.38, 0.23, 0.13, 0.05, 0, 0, 0.06, 0.14, 0.24,
+                    0.39, 0.5, 0.59, 0.68, 0.75]
+
+
+class Ddy(object):
```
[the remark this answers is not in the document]
chriswmackey: I thought we said that we were absolutely sure that we are following PEP8 here and naming all of the classes with title case.  The only calss that does not follow this right now is EPW.  The only reason why I haven't changed it is because I know that there should be a simultaneous update of the grasshopper components.  If you agree with me that we should just use title case, then I will update the EPW class and the Grasshopper components in my next pull request.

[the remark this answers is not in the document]
mostaphaRoudsari: I know we talked about this but looking back I think with acronyms we should keep them uppercase. That will make them easier to read and is closer to what a user would expect. I have seen it in Python libraries. Here is an example:

> class names: CamelCase, with acronyms kept uppercase (HTTPWriter and not HttpWriter)

from this page: https://palletsprojects.com/governance/sourcecode/python-styleguide/

tests/psychrometrics_test.py
chriswmackey: Done!

ladybug/designday.py
chriswmackey: I've now integrated all of these constants on the objects where they are used.  You are right that this is a lot cleaner.

ladybug/designday.py
chriswmackey: I was originally doing this for my own convenience since there are a number of workflows for creating design days without needing a .ddy file and it was useful to have a default.  I'll take this out and just force the user to specify a file path in these types of situations.

ladybug/designday.py
chriswmackey: You are right.  This is not needed as long as I have something to catch the case that no location object is found in the .ddy file.

ladybug/designday.py
chriswmackey: I spent a long time thinking about this.

The issue is that a ddy file can have only one location that is the same for all of the design days in the file.
However each design day has its own location, which is necessary for all of the solar methods that are on the design day object.
This creates the possibility in some workflows of someone creating a ddy file where all of the design days have different locations.  Therefore, we need a location property on the Ddy object that synchronizes all of the design days connected to the .ddy with a single location.

Yes, we can remove the default of creating a ddy file with no design days inside of it.

ladybug/designday.py
chriswmackey: I'm not sure if I follow the reasoning for why we should remove it.
Is there a reason why we shouldn't build in the convenience of saving a ddy file that has been edited with the API in the same folder as a ddy file where it originated from?

ladybug/designday.py
chriswmackey: Yes, I've put it as a property and referenced it in the doc string.

ladybug/designday.py
chriswmackey: I originally tried doing a regex but I could not come up with anything that was noticeably better or more reliable than these simple operations.  In the end, it seemed like the regex was overkill when all that I wanted to do was split the string and remove certain characters.

I should also note that the splitting and cleaning operations that I'm doing here can't result in a thrown exception and the check after this process makes sure that the data has been interpreted correctly (I'd need this check anyway whether a regex was used or not).

If you still think a regex is better, than I can do it but what I would implement is effectively the same as what I have done here (but with a longer calculation time that the regex uses).

Alternatively, if you can come up with a regex that it better or more reliable, we should definitely use it.

ladybug/designday.py
chriswmackey: Sorry.  This was mostly a brain fart. I have gone back and made sure all of these are correct.

ladybug/designday.py
chriswmackey: No. It has been removed.

ladybug/designday.py
chriswmackey: Done!

ladybug/designday.py
chriswmackey: I think you may be missing how design day temperatures work.  You can generate a list of hourly temperatures over the design day from the dry_bulb_max, the range over the day, and the temperature multipliers for each hour (which are standardized).  You can read here about how this works:
https://bigladdersoftware.com/epx/docs/8-9/engineering-reference/climate-calculations.html

ladybug/stat.py
chriswmackey: Done!

ladybug/designday.py
chriswmackey: Yes. Done!

ladybug/designday.py
chriswmackey: Ok. I added a setter that automatically converts the boolean values to "Yes" and "No" under the hood.

ladybug/designday.py
chriswmackey: Yep, that's cleaner!

ladybug/stat.py
```diff
@@ -1,10 +1,22 @@
...
+from .designday import WindCondition
+from .designday import RevisedClearSkyCondition
+from .designday import OriginalClearSkyCondition
 
 import os
 import re
 import codecs
 import platform
 
+months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul',
```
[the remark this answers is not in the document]
chriswmackey: I agree. Done!

ladybug/stat.py
chriswmackey: In the end, this turned out to be the same number of lines but I guess you are right that it's more readable.

chriswmackey: I'm also realizing that using regex here is making the code more prone to failure in the event that a given data type is not found in the .stat file.  Essentially, all of these if statements help guard in the event that there's missing data in the file.  If I want to make my regex code as resilient, I end up having to add back all of the if statements anyway to check and see if the regex successfully found a match.  Does this still sound worthwhile to you, @mostaphaRoudsari ?

ladybug/designday.py
chriswmackey: This has been addressed in our discussion

ladybug/designday.py
chriswmackey: This was addressed in our discussion

ladybug/designday.py
chriswmackey: I have implemented the regex as we discussed.  I agree that this way is much cleaner and more reliable than before.

ladybug/stat.py
chriswmackey: I implemented the regex as discussed

PR #71, opened 2018-09-16
Added EPW to json method

AntoineDao:
Added epw to_json() function to push epw data to db via web API. Have not created from_json() class method yet. Am happy to add it before merging this change if it seems more wise.

### DIFF
--- a/ladybug/epw.py
+++ b/ladybug/epw.py
@@ -293,6 +293,83 @@ class EPW(object):
 
         return full_path
 
+    def to_json(self):
+        epw_matrix = [
+            self.years.values,
+            self.dry_bulb_temperature.values,
+            self.dew_point_temperature.values,
+            self.relative_humidity.values,
+            self.atmospheric_station_pressure.values,
+            self.extraterrestrial_horizontal_radiation.values,
+            self.extraterrestrial_direct_normal_radiation.values,
+            self.horizontal_infrared_radiation_intensity.values,
+            self.global_horizontal_radiation.values,
+            self.direct_normal_radiation.values,
+            self.diffuse_horizontal_radiation.values,
+            self.global_horizontal_illuminance.values,
+            self.direct_normal_illuminance.values,
+            self.diffuse_horizontal_illuminance.values,
+            self.zenith_luminance.values,
+            self.wind_direction.values,
+            self.wind_speed.values,
+            self.total_sky_cover.values,
+            self.opaque_sky_cover.values,
+            self.visibility.values,
+            self.ceiling_height.values,
+            self.present_weather_observation.values,
+            self.present_weather_codes.values,
+            self.precipitable_water.values,
+            self.aerosol_optical_depth.values,
+            self.snow_depth.values,
+            self.days_since_last_snowfall.values,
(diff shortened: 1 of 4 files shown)


mostaphaRoudsari:
@AntoineDao, Thanks. What about epw header? Are you including the header into JSON data?

AntoineDao:
I knew I'd forget something! I used location information which I compose into the schema outside of ladybug. I could probably do so within actually... Maybe it will help to explain why I'm proposing this sort of change and find a better way to do it. Essentially I am wanting to relatively easily create a json output where epw hourly data is represented as a matrix (cols=type of data and rows=hours of the year) to facilitate data entry into a tabular data format (you guessed it an SQL database!).

AntoineDao:
Have changed according to discussion above. The json expression of an object should be as close as possible to the object it represent in python and manipulations to make that fit in an SQL database should be done by the API server (which it does!). The only cause for concern here is the `from_json` method because the EPW object relies on the order of a data collection on the `epw._data` list to index what type of data it is.

I reckon we should either change the way the EPW object store the data (ie: a dictionary) or we should find a way to make sure the data gets input in the correct order when performing `EPW.from_json()`.

ladybug/epw.py
```diff
@@ -87,16 +109,21 @@ def file_name(self):
...
+                    'Cannot find an epw file at {}'.format(self._file_path))
 
-        if not epw_file_path.lower().endswith('epw'):
-            raise TypeError(epw_file_path + ' is not an .epw file.')
+            if not epw_file_path.lower().endswith('epw'):
+                raise TypeError(epw_file_path + ' is not an .epw file.')
 
-        self._folder, self._file_name = os.path.split(self.file_path)
+            self._folder, self._file_name = os.path.split(self.file_path)
+        else:
```
mostaphaRoudsari: @AntoineDao, If the filepath is not set correctly we should raise an error, shouldn't we? Are you thinking about creating an EPW object from the data in database?

ladybug/epw.py
```diff
@@ -293,6 +320,12 @@ def save(self, folder=None, file_name=None):
 
         return full_path
 
+    def to_json(self):
+        return {
+            'location': self._location.to_json(),
+            'data': [data.to_json() for data in self._data]
+        }
```
mostaphaRoudsari: Still missing `self._header` otherwise it looks good to me!

mostaphaRoudsari:
@chriswmackey is going to make a number of changes to datacollection class which will affect epw and also epw header. Let's leave this out for now as EPW class will have several new properties that should be taken in consideration for to and from json.

AntoineDao:
Not gonna lie I had totally let go of this mentally! Happy to get back to it when data collections are updated etc...

chriswmackey:
I will also be parsing the header of the epw into a set of properties, which will change how stuff gets written to json as well. So, yes, let's pick this back up afterwards.

mostaphaRoudsari:
Thanks @AntoineDao! This is now addressed by @chriswmackey while refactoring datacollections. I'm going to close this PR.

PR #72, opened 2018-10-19
Small Fixes to DDY + STAT Objects and a Update to futil

chriswmackey:
This pull request is mostly cleaning up a few last things that were incomplete in my last commit including:
1) Additional use of Regex on the STAT object (both to enhance existing parsing and to parse in the typical / extreme weeks)
2) A cosmetic change to the DataCollection __repr__ to make it more readable as an object (and adding an isDataCollection property)
3) Added a function to download files, which works for all versions of Python (ironpython, python2 and python3).
4) Added some tests for the download file function to make sure that it is able to download from the Dept of Energy EPW database.

This commit is meant to be merged in at the same time as a merge into the ladybug-grasshopper reop.


ladybug/futil.py
```diff
@@ -1,6 +1,18 @@
 """A collection of auxiliary funtions for working with files and directories."""
 import os
 import shutil
+import zipfile
+try:
+    import System.Net as sys
```
mostaphaRoudsari: This should move to ladybug[+] libraries. We don't want to include any .Net stuff in core libraries.

https://github.com/ladybug-tools/ladybug-grasshopper/tree/master/ladybug

chriswmackey: I am not sure if I understand.  Are you saying that we need to duplicate this exact same download function in the plus library of each and every CAD interface (so one copy in the plus library for Dyanmo and another one in the Grasshopper plus)?  I'm just thinking that this might lead to the exact thing that we are trying to avoid.  If the internet security protocol gets updated again, we will have to make this change in each and every plus library currently in existence (which might have Blender, Sketchup, etc. in the future). 
I'll do it if you are confident that this shouldn't be here but a copy in each plus library didn't seem right to me.

mostaphaRoudsari: Dynamo and Grasshopper (which are the only .NET plugins) can have shared libraries if needed. SketchUp and Blender are Ruby and Cpython. I agree that we don't want to duplicate code but that should be addressed in its own way and is another issue to be discussed. In any case we don't want to have .NET dependencies in ladybug core libraries which is a CPython library.

To address that we might need to create a new library which is built on top of ladybug and is used by all .NET libraries.

ladybug-grasshopper : includes files that will be used but Grasshopper plugin
ladybug-net: includes .NET specific scripts which are shared between .NET plugins
ladybug: the core ladybug library

chriswmackey: Ah, ok.  This is clearer to me now and I agree that it does not belong in the core.  I didn't realize that the use of the .NET libraries would not be possible in Sketchup or Blender so this function isn't nearly as reusable as I thought.
Based on your comment, I can see that we might eventually need a separate ladybug-net repo to house functions like this but this might be a bit overkill for the time being.  For now, what do you think about me pulling this out into it's own dotnet.py file that will just be copy/pasted in the ladybug-grasshopper and ladybug-dynamo repos?  We can house this function in there for now and, if we start realizing that there are a lot of different functions that need to be there, we can make it it's own repo.

mostaphaRoudsari: >  For now, what do you think about me pulling this out into it's own dotnet.py file that will just be copy/pasted in the ladybug-grasshopper and ladybug-dynamo repos? We can house this function in there for now and, if we start realizing that there are a lot of different functions that need to be there, we can make it it's own repo.

Sure. Sounds like a plan!

ladybug/epw.py
```diff
@@ -241,57 +219,49 @@ def _get_data_by_field(self, field_number):
...
+                'saved as an EPW file.'
+            raise ValueError(length_error_msg)
+        else:
+            file_data = ''.join(lines)
+            write_to_file(file_path, file_data, True)
+        finally:
+            del(lines)
+            # move last item to start position for fields on the hour
+            for field in range(0, self._num_of_fields):
+                middle_hour = EPWFields.field_by_number(field).middle_hour
```
mostaphaRoudsari: Can you remind me what does `middle_hour` return? Is it the one that says if the data is shifted for an hour? why did we name it middle_hour?

ladybug/stat.py
```diff
@@ -191,13 +176,58 @@ def _import_data(self):
                 r"Monthly Statistics for Wind Speed[\s\S]*Daily Avg(.*)")
             for direction in self._wind_dir_names:
                 re_string = r"Monthly Wind Direction %[\s\S]*" + direction + r"\s(.*)"
-                self._monthly_wind_dirs.append(self._regex_parse(re_string))
+                dirs = self._regex_parse(re_string)
+                if dirs != []:
+                    self._monthly_wind_dirs.append(dirs)
+
+            # check to see if all monthly data is there
+            if self._monthly_db_range_50 != [] and self._monthly_wb_range_50 != [] \
```
mostaphaRoudsari: should this be a separate property? What if one changes the value for any of the lists and items? I'm not sure if that is a possibility but it seems that you want to wrap this in a separate method that gets called every time that you need to check if monthly value is found.

ladybug/stat.py
```diff
@@ -191,13 +176,58 @@ def _import_data(self):
...
+                    and self._monthly_wind != [] and self._monthly_wind_dirs != [] \
+                    and self._stand_press_at_elev is not None:
+                        self._monthly_found = True
         finally:
             statwin.close()
 
+    def _regex_check(self, regex_str, search_space, numbr=True):
+        matches = re.compile(regex_str).findall(search_space)
+        if len(matches) > 0:
+            if numbr is True:
```
mostaphaRoudsari: what if you try to cast to number and if it failed just return it as a string? Then you can remove the number argument. I'm thinking if there will be a case that one wants a numeric value not to be returned as a number.

ladybug/stat.py
```diff
@@ -156,15 +134,22 @@ def _import_data(self):
...
+            self._stand_press_at_elev = self._regex_check(
+                r"Elevation\s*[-]*\s*(\d*)Pa", self._header[5])
+            self._ashrae_climate_zone = self._regex_check(
+                r'Climate type\s"(\S*)"\s\(A', self._body, False)
+            self._koppen_climate_zone = self._regex_check(
+                r'Climate type\s"(\S*)"\s\(K', self._body, False)
+
+            # pull out extreme and seasonal weeks.
+            self._extreme_hot_week = self._regex_week_parse(
+                r"Extreme Hot Week Period selected:\s*(\S*)\s*(\S*):\s*(\S*)\s*(\S*),")
```
mostaphaRoudsari: We can probably be more specific in cases like this to avoid parsing errors. What you have there will also work of course.

```suggestion
                r"Extreme Hot Week Period selected:\s*(\w{3})\s*(\d{1,2}):\s*(\w{3})\s*(\d{1,2}),")
```

tests/epw_test.py
```diff
@@ -47,9 +47,23 @@ def test_import_tokyo_epw(self):
...
+        assert os.path.isfile(modified_path)
+
+    def test_save_wea(self):
+        """Test save wea_rel."""
+        path = './tests/epw/chicago.epw'
+        epw = EPW(path)
+
+        wea_path = './tests/wea/chicago_epw.wea'
+        epw.to_wea(wea_path)
+        assert os.path.isfile(wea_path)
```
mostaphaRoudsari: I would add checks to make sure the order of the data matched what is expected. This business of shifting data an hour back and forth can end up making issues in cases like this and adding a test will ensure that it works fine.

tests/futil_test.py
```diff
@@ -0,0 +1,48 @@
...
+        """Nothing to tear down as nothing gets written to file."""
+        pass
+
+    def test_download_file(self):
+        """Test the download file capability"""
+        wf_url = "https://www.energyplus.net/weather-download/" \
+            "southwest_pacific_wmo_region_5/AUS//AUS_NSW.Sydney.947670_IWEC/all"
+        wf_name = os.path.join("./tests/zip", "AUS_NSW.Sydney.947670_IWEC.zip")
+        futil.download_file(wf_url, wf_name)
+        assert os.path.isfile(wf_name)
```
mostaphaRoudsari: should you check approximate size? It can download an empty file.

tests/wea_test.py
```diff
@@ -136,6 +136,17 @@ def test_import_stat(self):
...
+    def test_write_wea(self):
+        """Test the write Wea file capability."""
+        stat_path = './tests/stat/chicago.stat'
+        wea_from_stat = Wea.from_stat_file(stat_path)
+
+        wea_path = './tests/wea/chicago_stat.wea'
+        hoys = range(8760)
+        wea_from_stat.write(wea_path, hoys, True)
+
+        assert os.path.isfile(wea_path)
```
mostaphaRoudsari: Again I would check the values in the wea file and the header in addition to if the file has been created.

ladybug/epw.py
chriswmackey: The middle_hour property tells you which of the EPW data types are recorded on the hour (when middle_hour is Flase) and which ones are represent a cumulative value of the previous hour (when middle_hour is True).  In this case it's being used to decide whether you need to take the last item of the EPW and put it in the spot to January 1 at 00:00.

ladybug/epw.py
```diff
@@ -60,9 +56,15 @@ class EPW(object):
...
-    def __init__(self, file_path=None):
+    def __init__(self, file_path):
         """Init class."""
-        self.file_path = file_path
+        self._file_path = os.path.normpath(file_path)
+        if not os.path.isfile(self._file_path):
+            raise ValueError(
+                'Cannot find an epw file at {}'.format(self._file_path))
+        if not file_path.lower().endswith('epw'):
+            raise TypeError(file_path + ' is not an .epw file.')
```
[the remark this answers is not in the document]
chriswmackey: Very good point.  Now that I understand format always casts to string, I'll be sure to use it here on out.

ladybug/stat.py
chriswmackey: Everything that it is checking is is a hidden attribute so I am not expecting people to set those properties (thereby needing a re-check).  If you think that some people might try setting the hidden attributes, I can make it it's own property.  It might just be a bit inefficient since it will get called every time someone calls one of the other properties instead of being called once but I can see this inefficiency would be worth it if people will be setting hidden attributes regularly.

ladybug/stat.py
chriswmackey: Ok.  I guess that this is a place where try / except is ok.  It gets a bit complicated for lists of numbers where one of the numbers is "N" indicating that there's no value for one of the items in the list.  But I think I can still write it to deal with this.

ladybug/stat.py
chriswmackey: Good idea.  I updated all of the regex conditions that were like this to use the same level of specificity.

tests/epw_test.py
chriswmackey: Good idea.  Thankfully, all of the radiation and illuminance data never gets shifted (so the Wea should be fine in this case).  Still, I'm adding it in now.

chriswmackey: Done

tests/futil_test.py
chriswmackey: Good idea.  Done!

tests/wea_test.py
chriswmackey: Done!

ladybug/stat.py
chriswmackey: I just implemented it as a property anyway.

chriswmackey:
@mostaphaRoudsari ,
I addressed all of your comments and I removed the extra files from the tests folder since you are right that they are not necessary.  I also added some code to clean up these files in the tests if we run them locally.  If you agree, I think that it is ready to be merged in now.

ladybug/epw.py
mostaphaRoudsari: Makes sense. I would still vote for renaming it to a more explicit name.

PR #81, opened 2018-11-07
fix(euclid3): enforce dependency version

### LINKED ISSUE #80: Iron-Python-check and stricter versioning for dependencies
Hi @mostaphaRoudsari , @chriswmackey  and @AntoineDao , 

Since dependencies are now being added through setup.py,  a few suggestions:
* Since IronPython 2.x is an incomplete port of Python 2.x, it might save some headaches later if there is some of testing to see if the dependency is not using some of the fancier features of Python 2 (xml is one of the modules that comes to mind).

* I highly recommend adding the version number to dependencies in `install_require`. For example, over [here](https://github.com/ladybug-tools/ladybug/blob/master/setup.py#L31-L33), it is safer if `euclid3` is `euclid3==0.01`. This isn't a big issue for euclid3 per se, but will likely be for more complex pacakges (I learnt this the hard way!).

AntoineDao:
Judging by the last deploy of euclid3 being 2014 I am not overly worried about breaking changes but this is a good habit to keep.

Should fix #80

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setuptools.setup(
     url="https://github.com/ladybug-tools/ladybug",
     packages=setuptools.find_packages(),
     install_requires=[
-        'euclid3'
+        'euclid3==0.1'
     ],
     classifiers=[
         "Programming Language :: Python :: 2.7",


AntoineDao:
Also, it's definitely `euclid3==0.1` as on the PyPi package it is marked as v0.01 (no dots!) =~ v0.1.

PR #86, opened 2018-11-24
ci(travis): split test and deploy jobs

AntoineDao:
Check out the Travis job to see the difference.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,11 +4,9 @@ python:
   - "2.7"
   - "3.6"
 
-env:
-  - TRAVIS=true
-
 install:
  - pip install pytest pytest-cov euclid3 coveralls
+
 
 script:
   - py.test --cov=. tests/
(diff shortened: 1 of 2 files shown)


AntoineDao:
Right. This should now work. It will only run the deploy job when a PR is merged into master branch or something is pushed directly to master.

.travis.yml
```diff
@@ -4,19 +4,25 @@ python:
...
-  - git config --global user.name "ladybugbot"
-  - git config --global user.email "release@ladybug.tools"
-  - pip install python-semantic-release
-  - semantic-release publish
+
+jobs:
+  include:
+    - stage: deploy
+      if: branch = master AND (NOT type IN (pull_request))
+      python: "3.6"
```
mostaphaRoudsari: @AntoineDao, Is this going to generate the whl file only for Python 3.6 or will it do it for both Python 2 and Python 3?

mostaphaRoudsari:
This looks much more cleaner. I checked the current release and the whl file is only generated for Python 2. 
https://pypi.org/project/lbt-ladybug/#files

I think with the new set-up it will be only generated for Python 3. How can we generate it for both Python 2 and 3? and does it even matter?

.travis.yml
AntoineDao: Short answer: I have no idea... Haven't dived that deep into python packaging. As far as I know it was initially only deploying using one python version anyways (either 2.7 or 3.6 depending on which finished testing and deploying first) and I have managed to run dragonfly on both python 2 and python 3 without any issues.

mostaphaRoudsari: from https://pythonwheels.com/

>Note: If your project is Python 2 and 3 compatible you can create a universal wheel distribution. Create a file called setup.cfg with the following content and upload your package.
```
[bdist_wheel]
universal = 1
```

AntoineDao: Nice!

PR #91, opened 2018-11-29
fix(wea): fix incorrect check for stat with no optical data

chriswmackey:


mostaphaRoudsari:
This is a typical item that we should have written a test for. Thank you for fixing it.

PR #96, opened 2018-12-21
feat(DataType): added datatype to handle unit conversions, checks, etc.

chriswmackey:
@mostaphaRoudsari ,

This is an initial PR before I take the plunge into restructuring the DataCollection class.  Specifically, this one does not yet introduce breaking changes and it fleshes out a complete datatype class that is stored on the DataCollection header and can be used both for unit conversions and checks.  Specifically, I tried to keep this PR to just tackle three issues:
* Implement a new datatype class that is linked to units and holds methods for unit conversion
* Add the datatype class to the header of the DataCollection
* Add methods to the DataCollection for to_ip(), to_si(), to_unit(), and is_in_range()

The new datatype (currently called datatypenew.py) has been added alongside the existing datatype.py class for now to avoid breaking changes.  In the next PR that I send, I will introduce breaking changes, renaming datatypenew.py to datatype.py and abolishing the existing datatype.y and the DataPoint class.

I apologize that the PR ended up being a bigger than initially anticipated because of two other things that these changes affected:
* The EPW class required some changes as I moved information to be stored in the DataType class instead of the EPWFileds class.
* As I was developing code that to allow easy conversion between units, it became apparent that we really should be distinguishing between 'radiation (Wh/m2)', which is a unit of energy intensity and is cumulative over time and '[irradiance (W/m2)](https://en.wikipedia.org/wiki/Solar_irradiance)', which is a unit of energy flux and is not cumulative.  Realizing that the Wea data is supposed to be irradiance (as we recently discovered by looking at Wea's from finer timesteps), I have changed the names of Wea properties to use irradiance instead of radiation.

One thing to note about this implementation of datatypenew.py is that I have made the possible units for a given base data type extensible so that we are not locked into one or two units per data type.  This not only allows us to use esoteric IP units (like refrigeration tons for cooling or knots for wind speed) but it also allows us to convert easily between different metric units.  For example, if I want to compare energy use results out of E+ (in J or GJ) with the rates that the utility charges for electricity (in kWh) and compare both of these to radiation benefit studies (in Wh). These types of conversions are now easy to do.

While I feel strongly that this flexibility of units is important (especially after 3 years of working in a US office), the one thing this does mean is that the to_si() and to_ip() aren't always going to a single unit.  But I feel that this is actually what we want for a lot of cases.  For example, I would like to have Wh converted to Btu and have kWh converted to kBtu.  I would also like to have miles of visibility converted to km and not inches or feet.  To deal with the fact that these to_ip() and to_si() methods aren't always guaranteed to return the same unit, I added to_unit() methods that allow you to pass the destination unit that you would like the DataCollection converted to and it will automatically change everything to be in that unit.

I will send a PR to the ladybug-grasshopper repo showing how I imagine these API calls being used within components.

I am happy to add more checks to the unit conversions if you approve the general concept here.

### DIFF
--- a/ladybug/datacollection.py
+++ b/ladybug/datacollection.py
@@ -54,19 +54,19 @@ class DataCollection(object):
         return cls(input_data, Header.from_json(data['header']))
 
     @classmethod
-    def from_list(cls, lst, location=None, data_type=None, unit=None,
-                  analysis_period=None):
+    def from_list(cls, lst, data_type=None, unit=None,
+                  analysis_period=None, location=None):
         """Create a data collection from a list.
 
         lst items can be DataPoint or numerical values.
 
         Args:
             lst: A list of data.
-            location: location data as a ladybug Location or location string
-                (Default: unknown).
-            data_type: Type of data (e.g. Temperature) (Default: unknown).
+            data_type: Type of data (e.g. Temperature).
             unit: data_type unit (Default: unknown).
             analysis_period: A Ladybug analysis period (Defualt: None)
+            location: location data as a ladybug Location or location string
+                (Default: unknown).
         """
         header = Header(data_type=data_type, unit=unit,
                         analysis_period=analysis_period,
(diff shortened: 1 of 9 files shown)


mostaphaRoudsari:
@chriswmackey thank you for the PR. Before I want to review the code I have a suggestion about having two very similar files.

> The new datatype (currently called datatypenew.py) has been added alongside the existing datatype.py class for now to avoid breaking changes. In the next PR that I send, I will introduce breaking changes, renaming datatypenew.py to datatype.py and abolishing the existing datatype.y and the DataPoint class.

What if you send this PR to the development branch and remove the old datatype class? We can keep the development in the development branch until we have all the other libraries as well as the plugins updated to work with the development branch. That will keep everything cleaner and avoid potential breaking changes in the plugins.

PR #101, opened 2018-12-29
feat(skymodel): Adding Perez split to the Zhang-Huang Model

chriswmackey:
This allows us to create solar radiation values when such data does not exist (which is for the vast majority of weather stations throughout the world). It will be instrumental in enabling a suite of functions for creating new EPW files from data taken from these stations.

Specifically, this PR augments the already-implemented Zhang-Huang solar model by incorporating a much more accurate way of splitting direct and diffuse solar.


mostaphaRoudsari:
Hey @chriswmackey, I think you meant to send this PR to `chris` branch so we just get the changes that you made after the previous PR. Right now it includes all the changes that you have made for both PRs.

image

chriswmackey:
Ok. I can close this out but remember that "chris" branch should be "datatype" branch and I thought that we are supposed to close out branches for a specific task once they are done.  Maybe I will just wait to send PRs until we have the first PR merged.

PR #103, opened 2019-01-02
Reorganizing Comfort Functions

chriswmackey:
This PR reorganizes the functions for thermal comfort in preparation for the new DataCollection class.


ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+                           low_bound=0, up_bound=100, error=0.001):
+    """Return the value of a missing_pmv_input given a target_pmv and the 6 other inputs.
+
+    This is particularly useful when trying to draw comfort polygons on charts
+    using the PMV model.
+
+    Args:
+        target_pmv: The target pmv that you are trying to produce from the inputs to
+            this pmv object.
+        other_inputs: A list of values for the other 6 inputs of the PMV model that
```
mostaphaRoudsari: I think there are better ways to handle this. You can be explicit about what you need and accept a dictionary. I would also suggest to use `tolerance` instead of `error`.

``` def calc_missing_pmv_input(target_pmv, input_values, ... , tolerance=0.001):```

Input values will be a dictionary with all the values. The missing key is your missing input.

File: ladybug/comfort/pmv.py [hunk 2]
```diff
@@ -1,1227 +1,555 @@
...
+            solution faster.
+        error: The acceptable error in the target_pmv. The default is set to 0.001
+
+    Returns:
+        missing_val: The value of the missing_input that will produce the target_pmv.
+
+    """
+    missing_pmv_input = missing_pmv_input.lower()
+
+    # Determine the function that should be used given the missing_pmv_i
```
mostaphaRoudsari: Can we change this to `None`?

ladybug/comfort/standard/comfortmodel.py
```diff
@@ -0,0 +1,24 @@
...
+
+class ComfortModel(object):
+    """Thermal comfort Model base class."""
+
+    def __init__(self):
+        self._calc_length = 0
+        self._is_computed = True
+
+    @property
+    def isComfortModel(self):
```
mostaphaRoudsari: This is a bad practice that I started to address the type issue in Dynamo. I think we should remove this from all the code base at some point.

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+
+    def thermal_condition(self, pmv, ppd):
+        """Determine whether conditions are cold, neutral or hot."""
+        if ppd >= self._ppd_thresh:
+            return 1 if pmv > 0 else -1
+        else:
+            return 0
+
+    def discomfort_reason(self, pmv, ppd, humidity_ratio):
+        """Determine if conditions are comfortable or not."""
```
mostaphaRoudsari: missing documentation. What does -2, 2 and 0 means.

It looks like the docstring for `thermal_condition` and `discomfort_reason` are swapped.

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+        """Initalize PMVComfortParameters
+
+        Properties:
+            ppd_comfort_thresh
+            humid_ratio_upper
+            humid_ratio_low
+            still_air_threshold
+        """
+
+        self._ppd_thresh = ppd_comfort_thresh or 10
```
mostaphaRoudsari: since 0 is an acceptable value for most of these inputs except for ppd you want to change them to also accept 0.

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+        self._hr_upper = humid_ratio_upper or 1
```
mostaphaRoudsari:
```suggestion
        self._hr_upper = humid_ratio_upper if humid_ratio_upper is not None else 1
```

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+        self._hr_lower = humid_ratio_lower or 0
+        self._still_thresh = still_air_threshold or 0.1
```
mostaphaRoudsari:
```suggestion
        self._still_thresh = still_air_threshold if still_air_threshold is not None else 0.1
```

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+
+
+class PMVComfortParameters(ComfortParameter):
+    """Parameters of PMV comfort."""
+
+    def __init__(self, ppd_comfort_thresh=None, humid_ratio_upper=None,
+                 humid_ratio_lower=None, still_air_threshold=None):
+        """Initalize PMVComfortParameters
+
+        Properties:
```
mostaphaRoudsari: Properties should go under Class docstring not __init__. Also the arguments description is missing.

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+class PMV(ComfortModel):
+    """PMV Comfort Object."""
+
+    def __init__(self, air_temperature, rel_humidity,
+                 rad_temperature=None, air_speed=None,
+                 met_rate=None, clo_value=None, external_work=None,
+                 comfort_parameters=None):
+        """Initialize a PMV comfort object from DataCollections of PMV inputs.
+
+        Args:
```
mostaphaRoudsari: Arguments description is missing!

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+
+        Returns:
+        """
+        raise NotImplementedError('PMV ComfortModel is not yet implemented.')
```
mostaphaRoudsari: I assume that this means that I shouldn't be reviewing this module. Is that correct?

tests/pmv_test.py
```diff
@@ -15,171 +21,61 @@ def tearDown(self):
...
-        wind_speed = 0.5
-        rel_humid = 60
-        met_rate = 1.5
-        clo = 0.4
-        external_work = 0
+    def test_pmv_fanger_utility(self):
+        """Test the pmv_fanger utility function"""
+        pmv_comf, ppd, hl = pmv_fanger(19, 23, 0.1, 60, 1.5, 0.4)
+        assert pmv_comf == pytest.approx(-0.680633, rel=1e-2)
+        assert ppd == pytest.approx(14.7373, rel=1e-2)
```
mostaphaRoudsari: Why not checking the value for `hl`?

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+
+    @property
+    def still_air_threshold(self):
+        """A number representing the wind speed beyond which the formula for Standard
+        Effective Temperature (SET) is used to dtermine PMV/PPD (as opposed to Fanger's
+        original equation). The default is set to 0.1 m/s.
+        """
+        return self._still_thresh
+
+    def is_comfortable(self, ppd, humidity_ratio):
```
mostaphaRoudsari: This is a method that I would expect to get from PMV class itself and not PMVParameters.

ladybug/comfort/standard/pmv.py
```diff
@@ -0,0 +1,406 @@
...
+    def ToString(self):
+        """Overwrite .NET ToString."""
+        return self.__repr__()
+
+    def __repr__(self):
+        """PMV comfort model representation."""
+        return "PMV Comfort Model: {} values".format(self._calc_length)
+
+
+class PMVComfortParameters(ComfortParameter):
```
mostaphaRoudsari: What if we change the name to `PMVParameters`? It is already under `comfort.standards`

tests/pmv_test.py
```diff
@@ -15,171 +21,61 @@ def tearDown(self):
...
-        pmv = pmv_comf.pmv
+        pmv_comf = PMVComfortParameters(
+            ppd_comfort_thresh, humid_ratio_up, humid_ratio_low, still_air_thresh)
+
+        assert pmv_comf.ppd_comfort_thresh == ppd_comfort_thresh
+        assert pmv_comf.humid_ratio_upper == humid_ratio_up
+        assert pmv_comf.humid_ratio_lower == humid_ratio_low
+        assert pmv_comf.still_air_threshold == still_air_thresh
+
+    def test_comfort_check(self):
```
mostaphaRoudsari: Missing tests for invalid inputs.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+        pmv, ppd, heat_loss = pmv_fanger(ta, tr, vel, rh, met, clo, wme)
+        ta_adj = ta
+        ce = 0
+    else:
+        ce_l = 0
+        ce_r = 40
+        eps = 0.001  # precision of ce
+
+        def fn(ce):
+            return (set - pierce_set(
```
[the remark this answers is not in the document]
chriswmackey: Fixed

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
     """
 
-    def __init__(self, air_temperature=None, rad_temperature=[], wind_speed=[],
-                 rel_humidity=[], met_rate=[], clo_value=[], external_work=[]):
-        """Initialize a PMV comfort object from lists of PMV inputs."""
-        # Assign all of the input values to the PMV comfort model object.
-        # And assign defaults if nothing has been connected.
-        self.air_temperature = air_temperature
+    result = {}
+    set = pierce_set(ta, tr, vel, rh, met, clo, wme)
```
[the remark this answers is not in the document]
chriswmackey: I changes this to `se_temp `to avoid this situation.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+        still_air_threshold=0.1):
+    """Calculate PMV using Fanger's original equation and Pierce SET model when necessary.
+
+    This method is the officially corrent way to calculate PMV comfort according to.
+    the 2015 ASHRAE-55 Thermal Comfort Standard.  This function will still return
+    accurate values even if the air speed is above the sill air threshold of
+    Fanger's original equation (0.1 m/s).
+
+    Args:
+        ta: air temperature [C]
```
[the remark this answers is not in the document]
chriswmackey: Everything has now been capitalized.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+            heat_loss: A list with the 6 heat loss terms of the PMV model [W].
+                The terms are ordered as follows:
+                    * heat loss by conduction
+                    * heat loss by sweating
+                    * heat loss by latent respiration
+                    * heat loss by dry respiration
+                    * heat loss by radiation
+                    * heat loss by convection
+
+    References
```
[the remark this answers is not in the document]
chriswmackey: All references have been moved to the correct location.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+    result['heat_loss'] = heat_loss
+
+    return result
+
+
+def pmv_fanger(ta, tr, vel, rh, met, clo, wme=0):
+    """Calculate PMV using only Fanger's original equation.
+
+    Note that Fanger's original expereiments were only conducted at
+    low air speeds (<0.1 m/s) and the pmv() function above should be
```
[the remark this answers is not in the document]
chriswmackey: Ok. I took out the reference to the location in the module and provided more background for cases where the pmv() method should be used.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+        heat_loss: A list with the 6 heat loss terms of the PMV model [W].
+            The terms are ordered as follows:
+                * heat loss by conduction
+                * heat loss by sweating
+                * heat loss by latent respiration
+                * heat loss by dry respiration
+                * heat loss by radiation
+                * heat loss by convection
+
+    References
```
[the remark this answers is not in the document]
chriswmackey: Done

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+        vel: relative air velocity [m/s]
+        rh: relative humidity [%]
+        met: metabolic rate [met]
+        clo: clothing [clo]
+        wme: external work [met], normally around 0 when seated
+
+    Returns:
+        pmv: predicted mean vote (PMV)
+        ppd: percentage of people dissatisfied (PPD) [%]
+        heat_loss: A list with the 6 heat loss terms of the PMV model [W].
```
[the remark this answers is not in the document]
chriswmackey: I would seriously consider taking this out, then, which would be a shame as I know several people are interested in this information.  Still, the outputs from this method really need to be small since this method will be run literally hundreds of thousands of times for an annual comfort map.  Having a dictionary with keys is going to significantly increase the memory requirements here. 

Or maybe, if I can't use a tuple or a list, I'll add another optional input to the method that allows you to request whether you want the heat loss terms or not?  Then I can output them as a dictionary.  Would that be better?

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
 
-        if rad_temperature != []:
-            self.__rad_temperature = rad_temperature
+    pa = rh * 10 * math.exp(16.6536 - 4030.183 / (ta + 235))
+
+    icl = 0.155 * clo  # thermal insulation of the clothing in M2K/W
+    m = met * 58.15  # metabolic rate in W/M2
+    w = wme * 58.15  # external work in W/M2
+    mw = m - w  # internal heat production in the human body
+    if (icl <= 0.078):
```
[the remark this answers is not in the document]
chriswmackey: Because that is a relic from when this code was originally written in javascript.  I have taken it out here.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+    hcf = 12.1 * math.sqrt(vel)
+    taa = ta + 273
+    tra = tr + 273
+    tcla = taa + (35.5 - ta) / (3.5 * icl + 0.1)
+
+    p1 = icl * fcl
+    p2 = p1 * 3.96
+    p3 = p1 * 100
+    p4 = p1 * taa
+    p5 = (308.7 - 0.028 * mw) + (p2 * math.pow(tra / 100, 4))
```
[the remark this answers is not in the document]
chriswmackey: Ok. Everything is a float now.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+        icl = 0.45
+
+    chc = 3.0 * pow(pressure_in_atmospheres, 0.53)
+    chcV = 8.600001 * pow((air_velocity * pressure_in_atmospheres), 0.53)
+    chc = max(chc, chcV)
+
+    # initial estimate of Tcl
+    chr = 4.7
+    ctc = chr + chc
+    RA = 1.0 / (facl * ctc)  # resistance of air layer to dry heat transfer
```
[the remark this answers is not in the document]
chriswmackey: Good point.  It's an old relic.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+    # Tcl and chr are solved iteratively using: H(Tsk - To) = ctc(Tcl - To),
+    # where H = 1/(Ra + Rcl) and Ra = 1/Facl*ctc
+
+    tcl_old = tcl
+    time = range(ltime)
+    flag = True
+    for TIM in time:
+        if flag is True:
+            while abs(tcl - tcl_old) > 0.01:
+                tcl_old = tcl
```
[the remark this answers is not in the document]
chriswmackey: That still needs to be set before the while loop somewhere.  How else can you have a while loop that is dependent on a variable that does not exist?

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+    # ========================  BEGIN ITERATION
+    #
+    # Tcl and chr are solved iteratively using: H(Tsk - To) = ctc(Tcl - To),
+    # where H = 1/(Ra + Rcl) and Ra = 1/Facl*ctc
+
+    tcl_old = tcl
+    time = range(ltime)
+    flag = True
+    for TIM in time:
+        if flag is True:
```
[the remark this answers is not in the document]
chriswmackey: Yes, this is a Javascript relic.  I moved the while loop out of the for loop.

ladybug/comfort/pmv.py
```diff
@@ -1,1227 +1,555 @@
...
+        pmv_hi = 3
+
+        def fn(pmv):
+            return (
+                (100.0 - 95.0 *
+                 math.exp(-0.03353 * pow(pmv, 4.0) - 0.2179 * pow(pmv, 2.0))) - ppd)
+
+        # Solve for the missing lower PMV value.
+        pmv_lower = secant(pmv_low, pmv_mid, fn, ppd_error)
+        if pmv_lower == 'NaN':
```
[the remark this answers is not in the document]
chriswmackey: Good point.  Javascript relic.

ladybug/comfort/pmv.py
chriswmackey: Ok. So we just expect a dictionary of pmv inputs that is missing one of the values?  We might have to do a lot more checks of the keys as a result but I can do that.

ladybug/comfort/pmv.py
chriswmackey: Change what to `None`?  The comment?

ladybug/comfort/standard/comfortmodel.py
chriswmackey: Ah, but I have written checks that look for this attribute in order to identify the type of input.  What is the correct way to check the type of object, then?

ladybug/comfort/standard/pmv.py
chriswmackey: Again, this class is not finished.  See comment above.

ladybug/comfort/standard/pmv.py
chriswmackey: Yes.

tests/pmv_test.py
chriswmackey: Well, let's first decide if I have to take it out based on comments above.

ladybug/comfort/standard/pmv.py
chriswmackey: There is no comfort model that actually tells you if people will be comfortable or if a given set of conditions are acceptable.  The PMV model only tells you what the mean vote should be across the population, UTCI only tells you the "feels like" temperature.  The judgement about whether a given level of PMV or UTCI are acceptable comes from committees and standards, such as the ASHRAE-55 standards (hence the name of this sub-module, standard).  These are wholly separate from the comfort models themselves and the scientists who made the models.  Furthermore, these standards different depending on building type (for example, some buildings have an acceptable PPD of 20% instead of the usual 10%).  These standards can also change from year to year depending on how the committee is feeling.  This is why this is separated out into a separate module.

ladybug/comfort/standard/pmv.py
chriswmackey: Good idea.

ladybug/comfort/standard/pmv.py
mostaphaRoudsari: Let me put it this way. When you calculate UTCI for a point in time and you want to know which condition of person this value represents where would you expect to find the method to evaluate that? Under UTCI class or under UTCI parameters? Just from API design point of view.

ladybug/comfort/pmv.py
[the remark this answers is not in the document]
mostaphaRoudsari: That's a fair point but doesn't justify removing it. In that case I think we should implement another version of the method which just returns minimum outputs that you will need for an annual comfort map and make the outputs for this one easier to access.

ladybug/comfort/pmv.py
chriswmackey: Ok. I have changed this over to a dictionary input with the result returning a dictionary that contains the missing key.

ladybug/comfort/standard/pmv.py
chriswmackey: Ah, wait. I see you were pointing to the other class. I just fixed this.

ladybug/comfort/standard/pmv.py
chriswmackey: I filled in some Arg docstrings anyway.

tests/pmv_test.py
chriswmackey: I added those checks.

ladybug/comfort/standard/pmv.py
chriswmackey: Yes, we have agreed that we will do it this way for now and, if it becomes confusing in the future, the methods for checking comfort can be moved to be standalone methods.  However, I don't think many people will want to use these methods since I imagine that most people will be trying to estimate thermal comfort for datacollections and there will be special objects to handle those.

ladybug/comfort/standard/comfortmodel.py
chriswmackey: We discussed that the isinstance() check is the right way to do this and we will plan to implement this in a later PR.

mostaphaRoudsari:
@chriswmackey, as discussed let's close this PR and you can send two separate PRs to ladybug and ladybug_comfort.

PR #105, opened 2019-01-09
ci(pytest): change pytest script and use dev-requirements.txt

AntoineDao:
CI and package versioning FTW!


chriswmackey:
Looks good @AntoineDao .  Thanks for fixing this.

PR #110, opened 2019-03-01
perf(datacollection): Refactored DataCollection class

mostaphaRoudsari:
BREAKING CHANGES: This is the complete set of changes to the datacollection module as we have discussed. The original DataCollection class has now been broken up into 5 classes and one hidden base class that they all inherit from.

Since the original datacollection.py file has changed dramatically, I think it's better to look at [the complete new file](https://github.com/ladybug-tools/ladybug/blob/datacollection-refactor/ladybug/datacollection.py) rather than the diff. You will also see some useful documentation at the top of the file describing the inheritance structure.

I am happy to report that the improvement in speed and memory usage across the library is substantial. Creating EPW objects now takes < 20% of the original time and many of the DataCollection methods also run in similarly small fractions of their original time. This can be chalked up to the fact that continuous hourly data collections now only create DateTime objects when they are needed and are not needed upon creation of the object. This means that, most of the time, we just rely on the start an end time of the analysis_period to figure out the datetime associated with values.

Note that I didn't make a specific collection for annual hourly data but rather made one for all forms of continuous hourly data. I did this because:

- The extra code and run time needed to make the methods work for all continuous data collections vs just annual ones is negligible.
- The improvements in speed and memory usage now apply to all energy simulation results, design day values, and all types of Weas, as well as EPW data.

Note that, as of the time that I am making this PR, I have not yet updated the tests or written new ones for the new DataCollections and I will push these tests shortly. I just wanted to get this up so that you have more time to review it.


AntoineDao:
I just noticed that the `BREAKING CHANGES` were not picked up by Semantic-Release :cry:. Is it worth making an empty commit delineating this change so laydbug can be bumped to 1.0.0? Also, is it worth running through committizen with you guys? I use it at work and it makes the whole commit message thing x100 times easier, especially when working with semantic release.

PR #113, opened 2019-03-02
feat(datatype): Add Data Types for Thermal Comfort

chriswmackey:
This is intended to be a small PR that adds a few extra data types necessary for thermal comfort calculations with data collections.

It also adds a new `compute_function_aligned` staticmethod to data collections after realizing that I was writing the same code over and over again in Grasshopper components that performed similar types of computations with data collections.  This staticmethod will prevent such repeated code from being written into UI components.

... and one other cleanup item: I removed the saturated_vapor_pressure_torr function from psychrometrics.py since it is used exclusively by the PMV function and now lives under ladybug_comfort.


ladybug/_datacollectionbase.py
```diff
@@ -407,6 +414,74 @@ def are_collections_aligned(data_collections, raise_exception=True):
...
+                                    'Got {}'.format(type(func_input)))
+
+        # run the function and return the result
+        if len(data_c) == 0:
+            # just inidivdual numerical values
+            return funct(*numerical_vals)
+        else:
+            # data collections
+            aligned_collections = []
+            for func_input in data_collections:
```
mostaphaRoudsari: Can't all this be done when you are doing the initial check in line 454?

It' not critical but performance-wise you will go through all the data once.

ladybug/_datacollectionbase.py
```diff
@@ -407,6 +414,74 @@ def are_collections_aligned(data_collections, raise_exception=True):
...
+                if isinstance(func_input, BaseCollection):
+                    aligned_collections.append(func_input)
+                else:
+                    aligned_collections.append(
+                        data_c[0].get_aligned_collection(float(func_input)))
+            BaseCollection.are_collections_aligned(aligned_collections)
```
mostaphaRoudsari: Do we really need this double check? since you just created the new `aligned_collections` it better be aligned! If there is a chance that they will end up not being aligned then we should get tit fixed when the issue happens.

ladybug/_datacollectionbase.py
```diff
@@ -407,6 +414,74 @@ def are_collections_aligned(data_collections, raise_exception=True):
...
+            result_values = []
+            for i in xrange(len(data_c[0].values)):
+                input_list = [coll[i] for coll in aligned_collections]
+                result_values.append(funct(*input_list))
+            return data_c[0].get_aligned_collection(
```
mostaphaRoudsari: I'm not sure why you have to create a list and then use another method to create a datacollection. Can't you just create a datacollection and add the values to it during the loop?

ladybug/datacollection.py
```diff
@@ -883,7 +884,13 @@ def get_aligned_collection(self, value=0, data_type=None, unit=None):
         else:
             data_type = self.header.data_type
             unit = unit or self.header.unit
-        values = [value] * len(self._values)
+        if isinstance(value, Iterable) and not isinstance(value, (str, dict)):
```
mostaphaRoudsari: You should check this but checking for `str` is not going to cover `bytes` and specially in Python3 one can send a byte which is iterable and is not str or dict but is not what you want. This comment is * 3 for other two similar cases in this PR.

ladybug/_datacollectionbase.py
chriswmackey: You were definitely right to point out that this could have been written a lot more elegantly and without the need to create a bunch of new lists (or new aligned data collections).  So I have cleaned it up a lot in the commit that I am about to push.  I found that there is a way to do it without any double iteration but it involves doing an `if` statement inside of a nested `for` loop like so:

`           result = data_colls[0].get_aligned_collection(data_type=data_type, unit=unit)
            for i in xrange(len(data_colls[0].values)):
                funct_input = []
                for col in data_collections:
                    funct_input.append(col if isinstance(col, float) else col[i])
                result[i] = funct(*funct_input)
            return result
`

I was curious if it would actually be faster so I profiled it and it turns out that the double iteration was a bit faster.  Granted, we are only talking about a difference of ~10ms in an operation that takes 40ms to run so it is nothing major.  But, given that it's basically the same number of lines of code and both of them seem to be equally readable, I am sticking with the double iteration.

ladybug/_datacollectionbase.py
chriswmackey: You are right that this was not the right place for the check.  However, it still needs to be there in case someone connects a list of two or more data collections that are not aligned with one another (otherwise, people will just get an IndexError instead of the more helpful one about non-aligned collections and which collections are problematic).  I have updated it so that it only runs this check for the data collections that are actually input to the method and not collections that are generated from individual float values.

ladybug/_datacollectionbase.py
chriswmackey: I don't have to do it this way and I guess you are right that the other way is more readable. So I have changed it over to be that way in the commit I am about to push.

However, both have similar performance because the current DataCollections don't allow you to append values.  Doing so would mean that you have to include a datetime for this new value or, in the case of a continuous collection, it would break the relationship between the number of values of the collection and the number of values permitted by the header analysis_period.  So, when I first create the collection, it actually comes with placeholder values that either have datetimes or have times that are fully described by the header analysis_period.  So, rather than appending values to the collection, I instead individually set the placeholder values to be the ones that I want.

ladybug/datacollection.py
chriswmackey: Thanks for letting me know.  This affects more places than just the 3 in this PR, including a number of places from the last PR where you brought up a similar concern and recommended the use of `isinstance(value, Iterable)` without acknowledging `bytes`.  So I am guessing that the discovery of the `bytes` object as an iterable might be relatively new for both of us.

I am going to fix all places where this happens now.

chriswmackey: Done!

chriswmackey:
Thanks @mostaphaRoudsari .  Everything should be addressed now.

PR #116, opened 2019-03-05
feat(datacollection): Consolidate Code

chriswmackey:
This PR consolidates the code that performs the different time interval operations on data collections (ie. average_monthly, total_daily, etc.)  It saves us from a lot of redundant code and also made it simple to implement an average_monthly, total_monthly, and percentile_monthly methods on the daily collection.

I also added a property on all data collections to track whether the data is_continuous.  This is obvious in the case of the hourly collections since there are two different objects but these methods are particularly helpful on the monthly, daily, and monthly_per_hour collections.  They essentially save us from having to make completely separate objects for continuous/discontinuous daily and monthly collections.

This PR also includes a few more changes to the datatype class in order for it to be able to handle thermal comfort conditions.  Specifically, I have moved the ThermalComfort datatype under the thermalcondition module where it seems to fit better.  I am also replacing the convention of unit_descr as a string with one as a dictionary.  This will make it easier to use this unit description to build ordinal legends when we start integrating datatype with legendpar.

### DIFF
--- a/ladybug/_datacollectionbase.py
+++ b/ladybug/_datacollectionbase.py
@@ -581,7 +581,7 @@ class BaseCollection(object):
         return d0 + d1
 
     def _average(self, vals):
-        return sum(vals)/len(vals)
+        return sum(vals) / len(vals)
 
     def _total(self, vals):
         return sum(vals)
@@ -612,7 +612,7 @@ class BaseCollection(object):
 
     @property
     def is_continuous(self):
-        """Return False."""
+        """Boolean denoting whether the data collection is continuous."""
         return False
 
     @property
--- a/ladybug/_datacollectionbase.py
+++ b/ladybug/_datacollectionbase.py
@@ -580,6 +580,17 @@ class BaseCollection(object):
         d1 = key(vals[int(c)]) * (k - f)
         return d0 + d1
 
+    def _average(self, vals):
+        return sum(vals)/len(vals)
+
+    def _total(self, vals):
+        return sum(vals)
+
+    def _get_percentile_function(self, percentile):
+        def percentile_function(vals):
+            return self._percentile(vals, percentile)
+        return percentile_function
+
     def __len__(self):
         return len(self._values)
 
@@ -598,6 +609,11 @@ class BaseCollection(object):
 
     def __contains__(self, item):
         return item in self._values
+
+    @property
+    def is_continuous(self):
+        """Return False."""
+        return False
 
     @property
     def isDataCollection(self):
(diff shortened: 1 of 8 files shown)


ladybug/_datacollectionbase.py
```diff
@@ -580,6 +580,17 @@ def _percentile(self, values, percent, key=lambda x: x):
         d1 = key(vals[int(c)]) * (k - f)
         return d0 + d1
 
+    def _average(self, vals):
+        return sum(vals)/len(vals)
```
mostaphaRoudsari:
```suggestion
        return sum(vals) / len(vals)
```

ladybug/_datacollectionbase.py
```diff
@@ -599,6 +610,11 @@ def __iter__(self):
     def __contains__(self, item):
         return item in self._values
 
+    @property
+    def is_continuous(self):
+        """Return False."""
```
mostaphaRoudsari:
```suggestion
        """Boolean denoting whether the data collection is continuous."""
```

ladybug/datacollection.py
```diff
@@ -550,6 +445,53 @@ def _check_analysis_period(self, analysis_period):
...
+        if interval == 'monthly':
+            data_dict = self.group_by_month()
+            dates = self.header.analysis_period.months_int
+        elif interval == 'daily':
+            data_dict = self.group_by_day()
+            dates = self.header.analysis_period.doys_int
+        elif interval == 'monthlyperhour':
+            data_dict = self.group_by_month_per_hour()
+            dates = self.header.analysis_period.months_per_hour
+
```
mostaphaRoudsari: I know this method is intended to be only used inside the class but it might be still useful to add an else and throw an error if interval is invalid.

```suggestion
            else:
                raise ValueError('Invalid input value for interval: {}'.format(interval))
```

ladybug/datacollection.py
```diff
@@ -967,6 +909,11 @@ def _get_analysis_period_subset(self, a_per):
         else:
             return AnalysisPeriod(*n_ap)
 
+    @property
+    def is_continuous(self):
+        """Return True."""
```
mostaphaRoudsari:
```suggestion
        """Boolean denoting whether the data collection is continuous."""
```

chriswmackey:
Thanks @mostaphaRoudsari !  I took all of your suggestions.

mostaphaRoudsari:
Thanks! After merging this it might be a good time to merge the development branch into master too.

chriswmackey:
Sounds good. I will set up the PR now.

chriswmackey:
I see that you already set it up.  I will check it out now.

PR #117, opened 2019-03-08
feat(datatype): Added data types for Specific Energy (Enthalpy) and Temperature Time (Degree Days)

chriswmackey:
Also, I started regretting that the data type for handling fractional values was written as 'percentage' instead of 'fraction.'  It seems that I was narrow-mindedly thinking of relative humidity in EPW files when I set up that type.  So I am fixing it here before we start having more dependencies on it.  Right now, it shouldn't really break anything outside of the API.


ladybug/datatype/fraction.py
```diff
@@ -49,72 +49,72 @@ def to_si(self, values, from_unit):
...
-class TotalSkyCover(Percentage):
+class TotalSkyCover(Fraction):
     # (used if Horizontal IR Intensity missing)
     _min = 0
-    _max = 100
+    _max = 1
     _abbreviation = 'CC'
     _min_epw = 0
-    _max_epw = 100
+    _max_epw = 1
```
mostaphaRoudsari: Isn't the value in epw between 1-10?

Change from % to fraction makes sense but shouldn't the max_epw values match the real values in epw file? I see that you are converting 100s to 1. How does this will play out with validating the EPW data itself?

Does it make sense to have % as a special type of fraction which is useful for cases like Relative Humidity and for the rest use the Fraction as needed.

ladybug/datatype/temperaturetime.py
```diff
@@ -0,0 +1,66 @@
+# coding=utf-8
+"""Temperature-Time data type."""
+from __future__ import division
+
+from .base import DataTypeBase
+
+
+class TemperatureTime(DataTypeBase):
```
mostaphaRoudsari: Didn't know it is called temperature time! Learned something new today.

ladybug/datatype/fraction.py
chriswmackey: The _max, _min, _max_epw and _min_epw properties are all recorded in the base units of the data type (in this case, fraction and not percentage).  The unit conversion from the base unit to whatever units the data collection is in happens when `is_in_epw_range()` or `is_in_data_type_range()` methods are called.  This way, if someone decides that they want to convert the cloud cover data to percentage or fraction and not tenths, the `is_in_epw_range()` and `is_in_data_type_range()` methods still work without having to track whether could cover is usually in tenths. Notably, the `is_in_epw_range()` and `is_in_data_type_range()` methods only convert the _min, _min, _max_epw and _min_epw properties to whatever the base units of the data collection are (we don't convert all of the values of the data collection).  So this makes such checks nice and efficient.

ladybug/datatype/temperaturetime.py
chriswmackey: I did some research and it seems that almost everyone calls it `degree-days` or `degree-hours` but such terminology confuses the name of the units with the name of the data type.  It would be like calling the `Temperature `data type `Degrees`, which will lead to a lot of confusion.  I settled on TemperatureTime after reading the [Wikipedia definition of degree-days](https://en.wikipedia.org/wiki/Degree_day):

A degree day is computed as the integral of a function of **time** that generally varies with **temperature**.

ladybug/datatype/fraction.py
mostaphaRoudsari: Hmmm. Thank you for the explanation. This is not a bug but one of those things that most likely goes under the `code smell` category. If we set up the min and max values for epw in epw fields then why do we have a duplicate property here which doesn't match the  epw? I _think_ if it is called min_epw or max_epw then it should follow min/ max in epw specification.

chriswmackey: I guess this is as good of a place as any to say that I have been seriously considering getting rid of the `_max_epw `and `_min_epw `properties for a number of reasons:

- Though these min and max values are written into the EPW documentation, they do not seem to be enforced anywhere.  EnergyPlus still runs fine with many of these limits exceeded in its input EPWs.
- Nearly every EPW throughout the DoE and OneBuilding databases violates at least one of these min/max limits (the most common one probably being illuminance though I have also seen it with relative humidity, wind speed, atmospheric pressure, and others).
- I have been struggling to find situations where people would actually want to run a check to see if data is in min/max epw range given that many real data sets exceed them.

So I have been considering removing these properties or at least removing the ability to check if data is within the epw range just because it is misleading.  I guess that we can still keep this min/max epw information on the data type objects for people's information but this also seems to not be clean.

In any case, I think the code smell that you are getting @mostaphaRoudsari  is not our own smell but rather the smell of writing arbitrary limits into the EPW documentation.  We have a case here where we can simply choose to ignore this situation just like EnergyPlus, and the DoE databases seem to do by not enforcing any of these limits.  Maybe we should follow their example.

chriswmackey: Now that I am getting the smell too, I think I would describe it as _contrived complexity_.  A simpler solution might just be to get rid of min_epw and max_epw if they are not useful.  And maybe the missing_epw property should actually be stored in the [epw_fields object](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/epw.py#L1386-L1642) since it only really relates to the EPW and not a more general definition of data type.

chriswmackey: Let me know if this sounds like a good solution and I am happy to implement it on this PR, @mostaphaRoudsari .

mostaphaRoudsari: I see your point. Let's add an issue and address the min/ max epw later. I can see use cases for this in the future but I think it should only live in one place which is most likely the EPW object. If we find the limitations doesn't make sense then we should contact Linda and Drew and get their input. Maybe we can use this as a case to update EPW documentation.

PR #119, opened 2019-03-09
fix(datatype): Putting epw missing, min, and max valules on EPW

### LINKED ISSUE #118: Get Rid of min_epw and max_epw Properties from DataTypes?
I have been seriously considering getting rid of the `_max_epw `and `_min_epw `properties for a number of reasons:

- Though these min and max values are written into the EPW documentation, they do not seem to be enforced anywhere.  EnergyPlus still runs fine with many of these limits exceeded in its input EPWs.
- Nearly every EPW throughout the DoE and OneBuilding databases violates at least one of these min/max limits (the most common one probably being illuminance though I have also seen it with relative humidity, wind speed, atmospheric pressure, and others).
- I have been struggling to find situations where people would actually want to run a check to see if data is in min/max epw range given that many real data sets exceed them.

So I have been considering removing these properties or at least removing the ability to check if data is within the epw range just because it is misleading given that the limits written into the epw documentation seem to be arbitrary.

A related change that I feel more strongly about is that the missing_epw property should actually be stored in the [epw_fields object](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/epw.py#L1386-L1642) since it only really relates to the EPW and not a more general definition of data type.  I could also see us moving the min/max epw properties here if we realize they are useful and not arbitrarily chosen.

chriswmackey:
After realizing that the code for the epw values on the data type objects did not pass the smell test, we are putting them back on the EPW field object.  I now agree that this makes more sense and we never should have moved them.

Resolves #118

### DIFF
--- a/ladybug/epw.py
+++ b/ladybug/epw.py
@@ -164,8 +164,8 @@ class EPW(object):
         # generate missing hourly data
         calc_length = len(analysis_period.datetimes)
         for field_number in xrange(6, epw_obj._num_of_fields):
-            data_type = EPWFields.field_by_number(field_number).name
-            mis_val = data_type.missing_epw if data_type.missing_epw is not None else 0
+            field = EPWFields.field_by_number(field_number)
+            mis_val = field.missing if field.missing is not None else 0
             for dt in xrange(calc_length):
                 epw_obj._data[field_number].append(mis_val)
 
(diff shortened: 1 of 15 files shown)


ladybug/epw.py
```diff
@@ -1425,150 +1425,201 @@ class EPWFields(object):
...
         8: {'name': fraction.RelativeHumidity(),
             'type': int,
-            'unit': '%'
+            'unit': '%',
+            'missing': 999,
+            'min': 0,
+            'max': 110
             },
 
         9: {'name': pressure.AtmosphericStationPressure(),
```
mostaphaRoudsari: @chriswmackey, out of curiosity why don't we hardcode the name of each field as a string?

chriswmackey: I am not sure that I follow your suggestion.  We ultimately need the data type object if we want to create data collections that can be converted to other unit systems and benefit from the other operations the data type objects offer.  Are you saying that we should store the name of the data type as a string and then use an exec() statement to get the data type object of the same name?  Or are you saying that we should have redundancy where the name exists as a sting on both the data type object and in this dictionary for epw fields?

PR #121, opened 2019-03-14
feat(datacollection): Added Immutable Types for Data Collections

chriswmackey:
I still have to finish the tests for the immutable types, which should be pushed shortly.

### DIFF
--- a/ladybug/_datacollectionbase.py
+++ b/ladybug/_datacollectionbase.py
@@ -173,7 +173,7 @@ class BaseCollection(object):
         """Get an immutable version of this collection."""
         if self._enumeration is None:
             self._get_mutable_enumeration()
-        col_obj = self._enumeration._immutable_collections[self._collection_type]
+        col_obj = self._enumeration['immutable'][self._collection_type]
         return col_obj(self.header, self.values, self.datetimes)
 
     def get_highest_values(self, count):
@@ -259,7 +259,7 @@ class BaseCollection(object):
         _filt_values, _filt_datetimes = self._filter_by_statement(statement)
         if self._enumeration is None:
             self._get_mutable_enumeration()
-        col_obj = self._enumeration._mutable_collections[self._collection_type]
+        col_obj = self._enumeration['mutable'][self._collection_type]
         collection = col_obj(self.header.duplicate(), _filt_values, _filt_datetimes)
         collection._validated_a_period = self._validated_a_period
         return collection
@@ -278,7 +278,7 @@ class BaseCollection(object):
         _filt_values, _filt_datetimes = self._filter_by_pattern(pattern)
         if self._enumeration is None:
             self._get_mutable_enumeration()
-        col_obj = self._enumeration._mutable_collections[self._collection_type]
+        col_obj = self._enumeration['mutable'][self._collection_type]
         collection = col_obj(self.header.duplicate(), _filt_values, _filt_datetimes)
         collection._validated_a_period = self._validated_a_period
         return collection
@@ -338,9 +338,9 @@ class BaseCollection(object):
             if self._enumeration is None:
                 self._get_mutable_enumeration()
             if mutable is False:
-                col_obj = self._enumeration._immutable_collections[self._collection_type]
+                col_obj = self._enumeration['immutable'][self._collection_type]
             else:
-                col_obj = self._enumeration._mutable_collections[self._collection_type]
+                col_obj = self._enumeration['mutable'][self._collection_type]
             collection = col_obj(header, values, self.datetimes)
         collection._validated_a_period = self._validated_a_period
         return collection
@@ -637,7 +637,16 @@ class BaseCollection(object):
         return percentile_function
 
     def _get_mutable_enumeration(self):
-        self._enumeration = _DataCollectionEnumeration(import_modules=False)
+        self._enumeration = {'mutable': {}, 'immutable': {}}
+        for clss in self._all_subclasses(BaseCollection):
+            if clss._mutable is True:
+                self._enumeration['mutable'][clss._collection_type] = clss
+            else:
+                self._enumeration['immutable'][clss._collection_type] = clss
+
+    def _all_subclasses(self, clss):
+        return set(clss.__subclasses__()).union(
+            [s for c in clss.__subclasses__() for s in self._all_subclasses(c)])
 
     def __len__(self):
         return len(self._values)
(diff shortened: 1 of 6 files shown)


chriswmackey:
@mostaphaRoudsari ,
This one is ready for your review whenever you get the chance.

ladybug/_datacollectionbase.py
```diff
@@ -256,8 +276,10 @@ def filter_by_pattern(self, pattern):
             A new Data Collection with filtered data
         """
         _filt_values, _filt_datetimes = self._filter_by_pattern(pattern)
-        collection = self.__class__(
-            self.header.duplicate(), _filt_values, _filt_datetimes)
+        if self._enumeration is None:
```
mostaphaRoudsari: why are we returning a mutable datacollection even if the original datacollection is immutable? shouldn't we just follow what the original object is?

ladybug/_datacollectionbase.py
```diff
@@ -623,3 +681,42 @@ def __repr__(self):
         """Discontinuous Collection representation."""
         return "Discontinuous Data Collection\n{} ({})\n...{} values...".format(
             self.header.data_type, self.header.unit, len(self._values))
+
+
+class _DataCollectionEnumeration(object):
```
mostaphaRoudsari: This is nice to have but can't we just hard-code these classes? unlike data-types we won't have so many datacollections that might be added to ladybug.

ladybug/datacollectionimmutable.py
```diff
@@ -0,0 +1,203 @@
...
+
+    @property
+    def values(self):
+        """The Data Collection's list of numerical values."""
+        return tuple(self._values)
+
+    @values.setter
+    def values(self, values):
+        try:
+            self._values
```
mostaphaRoudsari: a cleaner way is probably trying `hasattr(self, '_values')`

mostaphaRoudsari: you can probably simplify this as:

```python
if hasattr(self, '_values'):
   raise AttributeError(self._mutable_message())

self._check_values(values)
self._values = tuple(values)

```

ladybug/_datacollectionbase.py
```diff
@@ -586,6 +636,13 @@ def percentile_function(vals):
             return self._percentile(vals, percentile)
         return percentile_function
 
+    def _get_mutable_enumeration(self):
+        self._enumeration = _DataCollectionEnumeration(import_modules=False)
+
+    def _mutable_message(self):
```
mostaphaRoudsari: It works fine as it is but it might look cleaner to be set as a property. so you can get it as `self._mutable_message`.

ladybug/datacollectionimmutable.py
mostaphaRoudsari: same comments for all the classes in this module.

ladybug/_datacollectionbase.py
```diff
@@ -67,24 +73,17 @@ def header(self):
...
     def datetimes(self):
         """Return datetimes for this collection as a tuple."""
-        return tuple(self._datetimes)
+        return self._datetimes
 
     @property
     def values(self):
-        """Return the Data Collection's list of numerical values."""
+        """The Data Collection's list of numerical values."""
         return tuple(self._values)
```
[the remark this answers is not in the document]
chriswmackey: Good question.  So, even when the collection is mutable, we still need the result returned by `values` to be immutable because, if it returned a list, people would be able to append items to the list and then suddenly the number of values no longer matches the number of datetimes on the collection.  So I think we agree that the otuput here needs to be a tuple.

However, if we want to support people being able to set individual values of the collection by index by using `__setitem__`, then this gets very complicated when the underlying values are in a tuple and not a list.  Basically, every time that `__setitem__` gets called, we have to turn the underling values into a list, then change the single item to be the value that came through `__setitem__` and then we have to convert all of the values back to a tuple.  Such complication essentially destroys workflows where people initialize a Data Collection with dummy values and then set each item to be what they want.  If we want to get rid of the `__setitem__` capability, we can change the underlying values to be a tuple.  That way the only way people can edit a mutable collection is by using the values setter after building up a list of values.  This would make me a bit sad, though, since I really like the `__setitem__` workflow and I think that you do too.

ladybug/_datacollectionbase.py
chriswmackey: That was my original thought but then I realized what a mess this creates for all of the other methods that return data collections derived from a starting data collection.  For example, for all of the methods that return Daily, Monthly or MonthlyPerHour collections from hourly collections, do all of those also need to follow the same mutability as the starting collection?  If so, there will have to be code like the one you are commenting on here at least 20 times in datacollection.py since we will have to look up the mutable/immutable equivalent for Daily, Monthly or MonthlyPerHour collections on the hourly collection.

Then I asked myself, are there any cases I could think of where it is important that this mutability is preserved as one derives several new collections from a starting data collection. I could not think of any.  In fact, I would usually prefer to get a mutable version since it seems that the only cases where the immutable collections are important are the when they are tied to specific objects like the comfort collections.  So that is how I came up with these (hopefully pretty simple) set of rules that govern how immutability is conserved as one derives data collections from a starting collection:
https://github.com/chriswmackey/ladybug/blob/development/ladybug/datacollectionimmutable.py#L4-L12

ladybug/_datacollectionbase.py
chriswmackey: The difficulty there is that you will get circular imports of the data collections that inherit from one another.  Take, for example a standard mutable hourly collection. When you import this collection from datacollection.py, it does not import its immutable counterpart because its immutable counterpart inherits from the mutable one.  If we want to have a `to_immutable()` method on this object, there needs to be some way of importing the immutable one outside of the object itself.  I guess we could do this by copying a good fraction of this code that automatically imports all data collections into `__init__.py` if that sounds ok to you.

Then we could move the code that searches through all of the child classes of BaseCollection onto the BaseCollection object for all of the other classes to use.

Let me know if this sounds like the right solution to you and I can implement it.

ladybug/datacollectionimmutable.py
chriswmackey: Ahhhh!  That's much nicer.  Ok, let me do that.

chriswmackey: Man, that is so much better.  I was so used to using `hasattr` as an incorrect placeholder for `isinstance` that I forgot it has a real useful purpose!

ladybug/_datacollectionbase.py
chriswmackey: Also good idea.  Done!

chriswmackey:
@mostaphaRoudsari ,
I have responded to your comments and I am ready for feedback.

I just also wanted to let you know that I just discovered [multiple inheritance](https://www.python-course.eu/python3_multiple_inheritance.php) and realized that it could make some of the code for the immutable classes cleaner so I am just going to do a little reorganizing now.  This shouldn't have any bearing on the unresolved comments.

ladybug/_datacollectionbase.py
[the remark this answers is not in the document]
mostaphaRoudsari: > So, even when the collection is mutable, we still need values to return a tuple because, if it returned a list, people would be able to append items to the list

I see your point and making it a tuple make it harder but for what it worth you can add items to tuples without making it a list. try:  `(1, 2, 3) + (4, 5)`

ladybug/_datacollectionbase.py
mostaphaRoudsari: Thanks. Makes sense. I think the current implementation is good.  Let's keep it as it is.

ladybug/_datacollectionbase.py
mostaphaRoudsari: I think we both agree that this would be the expected behavior and I agree with you that it doesn't worth the effort at this point. Since it's well documented let's leave it as is and if someone wants to get started with contributing to the library this is a good one to take care of. We can always wrap this part of code in a utility function so we don't duplicate all the code. Thank you for the explanation.

PR #122, opened 2019-03-14
docs(sphinx): Added automated API documentation

mostaphaRoudsari:
We have been slow in updating the API documentation and as a result the API documentation is always behind the releases.

This PR will address this issue by automating the process of generating the documentation and deploying them to gh-pages. The documentation will only be generated from the master branch. If successful the API documentation should be accessible at: ladybug.tools/ladybug/docs 🤞


mostaphaRoudsari:
Thank you @chriswmackey! When do you think to merge the `development` branch into `master`? We should merge this PR before that and see how it will work in action. Once we know the workflow is working fine we can start fixing the warnings and document the issues as you said so we don't have similar warnings in the future.

@AntoineDao any comments on changes to `.travis.yaml` file?

mostaphaRoudsari:
There is no rush but let's push this in and see if it will work as expected once merged. We can always include the improvements later.

Issue #125, opened 2019-03-18
Deploy never gets to automated documentantion


mostaphaRoudsari:
The semantic release publication triggers a new built and that's the end of the build!

@AntoineDao has some ideas on how this can get fixed!

mostaphaRoudsari:
The issue is still there but for a different reason: https://travis-ci.org/ladybug-tools/ladybug/jobs/508786436

We currently have a similar issue for Honeybee too.

PR #126, opened 2019-03-19
fix(deploy): use javascript semantic-release package for deployment

AntoineDao:
Using the "pure" javascript implementation of semantic-release makes it a bit more flexible to set
up custom scripts to run at different stages of the semantic release process. The packages used as
well as scripts to run at different stages are specified in the .releaserc.json file. In this case
we use standard semantic-release and add a custom deploy script on top of the normal github release
setup. The custom deploy script is the new 'deploy.sh' file which takes the new version number as an
argument to deploy the package to PyPi and build the documentation.

#125


AntoineDao:
Finally made it after so many force pushes to get dev-requirements.txt to play ball!

PR #127, opened 2019-03-20
ci(travis): use python image to deploy with semantic-release

AntoineDao:
Seems it's easy enough to install node on a python image but the node image's python distro doesn't
play well with https.

#125


AntoineDao:
The job fails on my master branch at the semantic-release:deploy stage because the env vars are not set properly (wouldn't want me to deploy your packages would you? :stuck_out_tongue:

mostaphaRoudsari:
Merged! let's see how it will work. 🤞

PR #128, opened 2019-03-20
fix(deploy): fix travis file for docs deployment

AntoineDao:
#125 

Told you you'd be mad when you see the fix :stuck_out_tongue:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,11 @@ jobs:
     - git config --global user.email "releases@ladybug.tools"
     - git config --global user.name "ladybugbot"
     - npx semantic-release
-    provider: pages
-    skip_cleanup: true
-    github-token: $GH_TOKEN
-    keep-history: false
-    on:
-      branch: master
-    local_dir: docs/_build/
+    deploy:
+      provider: pages
+      skip_cleanup: true
+      github-token: $GH_TOKEN
+      keep-history: false
+      local_dir: docs/_build/
+      on:
+        branch: master
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 coverage==4.5.3
+coveralls==1.7.0
 euclid3==0.1
 pytest==4.3.1
 pytest-cov==2.6.1


AntoineDao:
Also, can we rebase and merge rather than use standard github merge system? It avoids having empty commits.

.travis.yml
```diff
@@ -30,10 +30,11 @@ jobs:
...
     - npx semantic-release
-    provider: pages
-    skip_cleanup: true
-    github-token: $GH_TOKEN
-    keep-history: false
-    on:
-      branch: master
-    local_dir: docs/_build/
\ No newline at end of file
+    deploy:
```
mostaphaRoudsari: do we need deploy twice? isn't the stage already set to deploy in line 20?

AntoineDao: Nah this is a specific marker to denote the deployment scripts in travis at any stage.

.travis.yml
```diff
@@ -30,10 +30,11 @@ jobs:
...
+      provider: pages
+      skip_cleanup: true
+      github-token: $GH_TOKEN
+      keep-history: false
+      local_dir: docs/_build/
+      on:
+        branch: master
```
mostaphaRoudsari: I missed this one! the `local_dir: docs/_build/` line is missing and as a result the documentation is gone! My bad. 😞 

@AntoineDao do you want to send a PR or do you want me to fix it?

mostaphaRoudsari: Oh! now I can see that it is not really missing. It's just up there! Does the order make a difference in this case?

AntoineDao: No it won't make a difference. However the latest build didn't run as expected because there were no changes in the commit history that would warrant a semantic relaese, therefore the docs were never built and an empty `docs` folder was pushed. Might have to change the logic here to support documentation building every time any commit is pushed.

AntoineDao: As in, this commit is the culprit: c08b7308077a48d5612f644943f92d5b5dade583 because as you can see [here](https://travis-ci.org/ladybug-tools/ladybug/jobs/509211619#L594) semantic release didn't run but [here](https://travis-ci.org/ladybug-tools/ladybug/jobs/509211619#L654) it deployed the boilerplate code  to gh-pages.

AntoineDao: I have one possible solution in mind: not generate the docs using semantic release every time, we should be able to configure a modules tree with 1 or two file configs rather than generating 100 different *.rst files for it.

mostaphaRoudsari: > Or we could generate the docs every time before we run any release script and get the version from the current latest tag in the git repo.

This is a good idea or the documentation should only happen when there is a change which means there is also a change in the semantic versioning. I think we should just set that up not to end up with empty docs otherwise the rest of set-up is pretty good if you ask me.

The 100 different *.rst files is intentional to keep the documentation in separate pages instead of one very long HTML page.

PR #130, opened 2019-04-12
 feat(skymodel): Added Perez Luminous Efficacy Model

chriswmackey:
This commit adds  the Perez model for luminaious efficacy of the sky, which seems to be the state-of-the-art model for estimating sky illuminance values from irradiance values.  More info on the model can be found in [this paper](http://www.cuepe.ch/html/biblio/pdf/perez-ineichen%201990%20-%20modelling%20daylight%20(se).pdf) and credit is due to @hansukyang for pointing me towards a Python implementation of this model.

With this commit, the Ladybug core now possesses enough climate models to make fully-simulate-able EPWs (for both daylight and energy) with only the following:
* dry bulb temperature
* dew point temperature
* atmospheric pressure
* sky cover
* wind speed
* wind direction


ladybug/skymodel.py
```diff
@@ -224,6 +224,129 @@ def zhang_huang_solar_split(altitudes, doys, cloud_cover, relative_humidity,
...
+            the kastenyoung1989 model to compute this value.
+
+    Returns:
+        gh_ill: Value for Global Horizontal Illuminance in lux.
+        dn_ill: Value for Direct Normal Illuminance in lux.
+        dh_ill: Value for Diffuse Horizontal Illuminance in lux.
+        z_lum: Value for Zenith Luminance in lux.
+
+    """
+    if altitude > 0:
```
mostaphaRoudsari: This is totally fine but it might have been easier to read if you would have wrote it the other way around:

```python
if altitude <= 0:
    return 0, 0 ,0, 0
# and the rest of the code goes here with no indentations
# ...
```

ladybug/wea.py
```diff
@@ -772,6 +801,42 @@ def write(self, file_path, hoys=None, write_hours=False):
 
         return file_path
 
+    @staticmethod
+    def _get_datetimes(timestep, is_leap_year):
```
mostaphaRoudsari: I know this is a left over from the code and you didn't add it in this PR but now that I see it we can probably remove this and only use `AnalysisPreiod(timestep=timestep, is_leap_year=is_leap_year).datetimes`

mostaphaRoudsari:
@chriswmackey should we merge in this PR?

ladybug/skymodel.py
chriswmackey: Good point. Done!

ladybug/wea.py
chriswmackey: I wish it could be that simple but, unfortunately, we need a special formula here to deal with the fact that the EPW datetimes for radiation fall on the half-hour while, at finer timesteps, we expect the datetimes to start counting from 0:00 on Jan 1 and not 0:30. We could still use the AnalysisPeriod object to help us here but it's really only saving us a line of code and it comes at the expense of being slightly longer to run. So, I think it is best to leave it as it is for now.

chriswmackey:
Yes, @mostaphaRoudsari . Sorry it took so long for me to respond to your comments. I'll merge it in now.

PR #139, opened 2019-04-30
docs(conf.py): set version in conf.py

mostaphaRoudsari:
This should fix the automated documentation.


AntoineDao:
Yep that totally worked :smile:

PR #141, opened 2019-05-30
perf(geometry): Replacing euclid with ladybug_geometry

chriswmackey:
This PR replaces the euclid dependency in ladybug core with ladybug_geometry.

I also use ladybug_geometry to start "solving" the visualization issue that, up until this point, has been unaddressed.  Specifically, I finished a LegendParameter and Legend class that deal with legend visualization by generating ladybug geometires.  I have written some code over in the ladybug_rhino repository that translates this to Rhino and I have done a lot of visual tests producing legends like this:
image
... and this:
image
I can add some more tests to this PR soon but, given that the new code is only 100-200 lines, I'm just sending it as an official PR


AntoineDao:
Changes look ace @chriswmackey! Can we clean up the branch commit history before merging though? I can give you a hand if you're not too sure how to go about it :smile:

chriswmackey:
Thanks @AntoineDao . @mostaphaRoudsari introduced me to the beauty of GitKraken and it should be a lot cleaner now.

ladybug/color.py
```diff
@@ -322,132 +313,113 @@ class ColorRange(object):
...
-        colorRange = ColorRange(chart_type = 1)
+        colorRange = ColorRange(continuous_colors=False)
         colorRange.domain = [100, 2000]
         colorRange.colors = [Color(75, 107, 169), Color(245, 239, 103),
             Color(234, 38, 0)]
         colorRange.color(300)
         >> <R:245, G:239, B:103>
-
-        ##
-        colorRange = ColorRange(chart_type = 2)
```
mostaphaRoudsari: @chriswmackey before I start to review am I understand this correctly that you are removing ordinal charts? and if yes why are we doing that?

chriswmackey: @mostaphaRoudsari ,

I removed this only from the ColorRange object and I put the ability to create ordinal legends on the LegendPar and Legend classes through a dictionary that maps values to text categories:
https://github.com/ladybug-tools/ladybug/pull/141/files#diff-600acdaadb3713444a5c0a272f102834R285

My reasons for this are several-fold:

1) I know that you originally were thinking that things like data collections would have text data in them but, at this point, all data collections and data types are essentially numerical, though some data types [carry dictionaries that map values to text categories as you see here](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/datatype/thermalcondition.py#L46-L47).  These dictionaries are essentially what will go into the LegendParameter ordinal_dictionary by default when no one has been specified.

2) With the way I have implemented the ordinal_dictionary now, users are forced to relate the text categories to numbers, making it possible to tell what the order of the text categories should be.  This was an issue before since it was difficult to tell which categories were "smaller" or "larger" than others. Having a dictionary like this also keeps things flexible if one decides that they don't want to display the full range of text categories that is available in the dictionary (ie. you only care about a PMV between -1 and +1 and not -3 and +3).  If you wanted to do this previously, you would have to edit the domain of the color range AND the values that you sent to the color range.  Now, you just change the min and/or max of the LegendParameters and it should all work.

3) If all that you really need is an object that maps text categories to colors, you don't need a whole ColorRange object capable of calculating unique colors for numerical values. Just use a dictionary.  It's just not worth it to add all of these extra `if` statements to the ColorRange object just to make it capable of doing something that a simple dictionary would do better.

mostaphaRoudsari: Two questions:

1. How does this approach will work for non-datatype data?
2. How would we map range of data to ordinal values? For instance < 300 ('low'), between 300-2000 ('desired') and > 2000 as ('too much')? Can we handle that?

mostaphaRoudsari: PS: Also see d3js scale categories:

https://github.com/d3/d3-scale#api-reference

chriswmackey: @mostaphaRoudsari ,

To answer your questions:

1. I am not sure I understand the question since the legend and color range exist independently of the datatype module.  They are built to take a set of any numerical data and make colors/ legends with it. So, if you want your legend to display with text values, and you don't have an ordinal dictionary from a DataType, you just make your own ordinal dictionary and assign it to the `ordinal_dictionary` of the LegendParameters.  We can have a component or a special way of dealing with Grasshopper inputs to help people make these dictionaries.

2. I am not sure if you are asking how to make a legend that displays with these text categories or if you are asking how can we produce colors that fall into 3 categories from a continuous data set.  So I will answer both. The first one I described above and in which case you'd make a legend with 3 segments at min/max values of 300 and 2000 and an ordinal_dictionary like so `{300: 'low', 1150: 'desired', 2000: 'too much'}`.  You can also make it with 5 segments and you will just get blank values for two of the segments in between the categories. For the second, you will see the [color range still has an option for continuous or discrete colors](https://github.com/ladybug-tools/ladybug/pull/141/files#diff-604a419b183a1c82099ebc5480cb0241R324).  The default is the continuous version we are used to but, if you only want to produce only 3 colors from the ColorRange, you just input 3 colors to the ColorRange and set the `continuous ` to False.

mostaphaRoudsari: Thanks.

1. Makes sense and it is a good approach.

2. I don't really follow all the details here. For instance I'm not sure why did you put 1150 there as a threshold but if I understand correctly it is possible to do so. Since this will be a common case I think it is helpful to add a sample code in tests to explain the workflow and also ensure it does work as expected.

chriswmackey: 2. I put 1150 there because it is the midpoint between 300 and 2000 and it is a linear scale.  I will add some code in the tests that demonstrate how to do this and I will adapt the most useful examples for the `Usage:` part of the docstring.  

The point I made above does bring up one limitation of what I have implemented now, which is that the CorlorRanges is assumed to be a linear scale (it is not possible to do exponential or logarithmic scales now).  For now, I can't really think of cases where we really want an exponential or logarithmic scale but, if we want to add this in the future, the presumption would be that we add a special type of color range object that accepts the exponent and offset of a power scale (or vice versa for the log scale).  We can then add an option to the LegendParameters that changes the base color scale to one of these objects.

mostaphaRoudsari: >  For now, I can't really think of cases where we really want an exponential or logarithmic scale but, if we want to add this in the future, the presumption would be that we add a special type of color range object that accepts the exponent and offset of a power scale (or vice versa for the log scale). We can then add an option to the LegendParameters that changes the base color scale to one of these objects.

Or we can just keep ordinal option which gives you full freedom for such cases. There are several cases that the values are not linear and having an ordinal legend makes it quite easy to handle such cases. That is why I put it in legend in the first place.

chriswmackey: I think you are confusing the generation of colors for a set of values with the display of the legend.  You can create a legend that displays exponential values with the ordinal_dictionary.  I was talking about the generation of colors according to a power or log scale (something that was not there in the previous ColorRange and is not there now).

The ordinal legend that you originally had would not let you set up an power or log color scale because you need to let the user specify the exponent and offset if you are actually going to enable this.  Your original scale (and the one there now) lets you approximate a power or log scale by plugging in a list of sequentially exponential values for the domain but the interpolation between the values of each domain member is still linear. So it's not a true exponential or log scale.

Maybe I should just hold off on giving more of a description until you have the time to further review the code as it sounds like I create more confusion than clarity when I explain things.

chriswmackey: ... or perhaps what will make it clearest is if I add some `Usage` cases.  I'll do that shortly.

chriswmackey: I have added a couple of resources (noted in the reply below) that should make things much clearer.

chriswmackey:
Ok, I have added two things that should make it much more clearer that anything that I could explain with words here. The first is that I added a `Usage` section to the docstring showing how the legend works:
https://github.com/chriswmackey/ladybug/blob/development/ladybug/legend.py#L39-L90

The second is that I put together a Grasshopper definition that shows how things will work visually, which, given the fact that this is a visual type of PR, will be much more helpful than any verbal description that I will give here:
https://drive.google.com/file/d/1F6WoqYYZD09v9o1XB45wTU6VeAgLPUBW/view?usp=sharing

You will see that, to run it, you need the copy of ladybug core in my fork as well as `ladybug-geometry`, `ladybug-rhino`, and `ladybug-dotnet`.  So I know it takes some time to set up but I think this might be much more helpful for showing how things work instead of the code here

ladybug/psychrometrics.py
```diff
@@ -6,11 +6,17 @@
 
 
 def saturated_vapor_pressure(t_kelvin):
-    """Saturated Vapor Pressure (Pa) at t_kelvin (K).
+    """Saturated vapor pressure (Pa) at a given dry bulb temperture (K).
```
mostaphaRoudsari: @chriswmackey can you keep your changes in a different branch so we keep this PR focused on what it is supposed to be? Once this is merged you can merge your new changes and send a new PR.

mostaphaRoudsari: PS: I will review this PR later tonight.

chriswmackey: Ok. Sorry about that. I was riding the wave of all the tests that I was writing and I started writing them for psychrometrics without realizing how how far off base I was from the original PR. I will continue in another branch.

ladybug/color.py
mostaphaRoudsari: Thank you for posting the sample code. The second sample shows that it is possible to do what I was looking for. I still think the set-up for ordinal dict is not clean. You have to come up with 1150 which is not really a number of interest. I get the logic of how you came up with this number but from user point of view having an explicit domain of 300 to 2000 is easier to understand and what I expect to put in if that is the range. In any case, since it works we can keep it as is.

I leave this part of the code a reference.

```
        data = [100, 300, 500, 1000, 2000, 3000]
        ordinal_dict = {300: 'low', 1150: 'desired', 2000: 'too much'}
        legend = Legend(data, LegendParameters(min=300, max=2000, number_of_segments=3,
                                               ordinal_dictionary=ordinal_dict))
        print(legend.segment_colors)
        legend.legend_parameters.continuous_colors = False  # get data in only 3 colors
        print(legend.value_colors)  # data colors align with number_of_segments
        >> ((R:75, G:107, B:169), (R:115, G:147, B:202), (R:115, G:147, B:202),
            (R:115, G:147, B:202), (R:115, G:147, B:202), (R:234, G:38, B:0))
```

ladybug/psychrometrics.py
mostaphaRoudsari: Thanks!

ladybug/legend.py
```diff
@@ -0,0 +1,877 @@
...
+    @property
+    def values(self):
+        """The data set assigned to the legend."""
+        return self._values
+
+    @property
+    def value_colors(self):
+        """A List of colors associated with the assigned values."""
+        _color_range = self.color_range
+        return tuple(_color_range.color(val) for val in self.values)
```
mostaphaRoudsari: Why not a list?

ladybug/legend.py
```diff
@@ -0,0 +1,877 @@
...
+
+        is_number_of_segments_default
+        is_title_default
+        is_base_plane_default
+        is_segment_height_default
+        is_segment_width_default
+        is_text_height_default
+    """
+
+    def __init__(self, min=None, max=None, number_of_segments=None,
```
mostaphaRoudsari: This is fine for this case but this is too many input arguments. Is it possible to bundle some of them? Do you think one will need to input all these values when the legend parameters is initialized or we can save some of them to be set as properties if needed. Can we set some of Boolean inputs as default and then user can overwrite them if they needed to do so?

```python
lp = LegendParameters(min, max, seg_count, colors)
lp.horizontal = True
lp.text_height = 5
```

ladybug/legend.py
```diff
@@ -0,0 +1,877 @@
...
+            assert isinstance(maximum, (float, int)), \
+                'Expected number for max. Got {}.'.format(type(maximum))
+            if self._min is not None:
+                assert maximum >= self._min, \
+                    'Input max is less than input min. {} < {}.'.format(
+                        maximum, self._min)
+        self._max = maximum
+
+    @property
+    def number_of_segments(self):
```
mostaphaRoudsari: I originally used `number_of_x` for a lot of cases but came to realize `x_count` is shorter and started using that. We should probably use one of them to keep the code consistent.

ladybug/legend.py
```diff
@@ -0,0 +1,877 @@
...
+            self._number_decimal_places = 2
+
+    @property
+    def include_larger_smaller(self):
+        """Boolean noting whether > and < should be included in legend segment text."""
+        return self._include_larger_smaller
+
+    @include_larger_smaller.setter
+    def include_larger_smaller(self, lgsm):
+        if lgsm is not None:
```
mostaphaRoudsari: I saw this a couple of times and didn't leave a comment but just as suggestion I would say just a quick check is enough for a Boolean.

```
self._include_larger_smaller = bool(lgsm)
```

ladybug/legend.py
```diff
@@ -0,0 +1,877 @@
...
+        min = self.min if self.min is not None else '[default]'
+        max = self.max if self.max is not None else '[default]'
+        seg = '[default]' if self.is_number_of_segments_default \
+            else self.number_of_segments
+        title = '[default]' if self.is_title_default else self.title
+        base_pt = '[default]' if self.is_base_plane_default else self.base_plane.o
+        seg_h = '[default]' if self.is_segment_height_default else self.segment_height
+        seg_w = '[default]' if self.is_segment_width_default else self.segment_width
+        txt_h = '[default]' if self.is_text_height_default else self.text_height
+        return 'Legend Parameters\n minimum: {}\n maximum: {}\n segments: {}\n' \
```
mostaphaRoudsari: Might be too much info for a quick print looks like a debugging output. If user needs to access a certain property they can just access that.

ladybug/resultmesh.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Args:
+            values: A List or Tuple of numerical values that will be used to
+                generate the legend and colors.
+            mesh: A Mesh3D object, with a number of faces or vertices that
+                match the number of input values and will be colored with
+                restults.
+            legend_parameters: An Optional LegendParameter object to override
+                default parameters of the legend.
+        """
```
mostaphaRoudsari: docstring for unit is missing.

ladybug/resultmesh.py
```diff
@@ -0,0 +1,223 @@
...
+        # check the inputs
+        assert isinstance(mesh, Mesh3D), \
```
mostaphaRoudsari: This can also be useful for other visualizations like sun-path with a legend. Maybe we should ask for a bounding box more than a Mesh input.

Also I feel this class can be a method to give suggestive values. Right now the logic is built-in and so making changes to it is hard. I can see it as a function that takes a geometry and legend parameters and returns the legend with the suggested location which user can modify before pushing it for visualization.

ladybug/color.py
```diff
@@ -268,186 +322,192 @@ def therm(cls):
...
+        if self._continuous_colors:  # continuous
+            # if type is continuous domain can only be 2 values
+            # or at least 1 value less than number of colors
+            if len(dom) == 2:
+                # remap domain based on colors
+                _step = float(dom[1] - dom[0]) / (len(self._colors) - 1)
+                _n = dom[0]
+                dom = tuple(_n + c * _step for c in range(len(self._colors)))
+
+            assert len(self._colors) >= len(dom), \
```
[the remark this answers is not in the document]
chriswmackey: You can see [here](https://github.com/ladybug-tools/ladybug/pull/141/files/dfce249ef88d2f4104e55a53ecc32884edd437fa#diff-604a419b183a1c82099ebc5480cb0241L390) that this whole section was from your original code and so that is why it's here now.  I agree that it doesn't make much sense and an `else` statement would be better.  I'll add it now.

ladybug/color.py
```diff
@@ -488,12 +561,20 @@ def __getitem__(self, key):
...
-        self._colors[key] = value
-
-    def __delitem__(self, key):
-        """Remove a color from the color list."""
-        del self._colors[key]
+        assert isinstance(value, Color), 'Expected ladybug color. ' \
+            'Got {}'.format(type(value))
+        _new_cols = list(self._colors)
+        _new_cols[key] = value
+        self._colors = tuple(_new_cols)
```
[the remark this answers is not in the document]
chriswmackey: Are we not worried about the case where someone appends or deletes a color from the list and then breaks the ColorRange because we were not able to perform a check to make sure the number of colors aligns with the number of domain items (what happens now in the `colors.setter`)? I understand why we want to avoid casting everything to a tuple in some cases but the color range gets used so infrequently that it seems it is worth an extra casting for the peace of mind that the `colros` property is valid.

ladybug/legend.py
```diff
@@ -0,0 +1,877 @@
...
+        # check the inputs
+        assert isinstance(values, Iterable) \
+            and not isinstance(values, (str, dict, bytes, bytearray)), \
+            'values should be a list or tuple. Got {}'.format(type(values))
+        self._values = values
+        if legend_parameters is not None:
+            assert isinstance(legend_parameters, LegendParameters), \
+                'Expected LegendParameters. Got {}.'.format(type(legend_parameters))
+            self._legend_par = legend_parameters.duplicate()
+            self._is_legend_parameters_default = False
```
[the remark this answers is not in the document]
chriswmackey: This is helpful for the objects that possess the `Legend` as one of their properties (like the `ResultMesh` object).  Particularly when we call the `to_dict()` method, we know that, if the `LegendParameters` assigned are default, there's no need to write them into the dictionary (the default ones will just be re-assigned when the object is re-serialized).  Admittedly, this is not a critical feature and everything will still work if we always write explicit `LegendParameters` into the dictionary. However, if our goal with the dictionary serialization is to keep the dictionary description of the object as minimal as is necessary, this seems like the right way to do it.

ladybug/legend.py
chriswmackey: You are right that there isn't much harm here in making it a list since it's always regenerated every time the property is called (so it won't create bad mutations of the object).  I guess it comes down to whether any benefits of using a list outweigh the extra memory a list has over a tuple and the fact that people will be able to make it out-of-sync with the values assigned to the object.  So what are the benefits of using a list here?

ladybug/legend.py
chriswmackey: You can see that all of the inputs to the `LegendParameters` have defaults (specifying None or leaving them blank essentially sets a default).  Also, all of the properties have setters, meaning there is no pressure to specify anything on `__init__`.
So the code you have written there will work (with the exception that the `horizontal` property is called `vertical_or_horizontal`).
We could bundle some of the properties into another object but I couldn't think of any properties that could be done so neatly or intuitively. In most cases, it felt like it was adding extra complexity.  If you have a suggestion for how to group them, let me know.

ladybug/legend.py
chriswmackey: Took me some time to realize what you meant by `x` but now I see that you were just saying that it's a general rule you use across all property names with that structure.  Ok, I will change it to `segment_count`

chriswmackey: I'll do the same thing with `number_decimal_places`. `decimal_count` is much better.

ladybug/legend.py
chriswmackey: Noted and implemented.

ladybug/legend.py
chriswmackey: I originally had it much leaner but I found myself constantly wanting to check defaults that were being assigned for all the properties, especially from Grasshopper.  I guess we can just have the ToString() method output this text and make the `__repr__` lighter. Does that sound like a better alternative and, if so, should the `__repr__` contain any of the properties?

ladybug/resultmesh.py
chriswmackey: Good catch.

ladybug/color.py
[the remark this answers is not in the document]
mostaphaRoudsari: Makes sense but in that case do we really need to support item assignment in the first place?

ladybug/legend.py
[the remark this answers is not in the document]
mostaphaRoudsari: Generally speaking the logic of keeping it as small as possible makes sense. Since we won't send 1000s of LegendParameters I would say in this case being explicit is the better approach and the size is not the critical factor.

ladybug/legend.py
mostaphaRoudsari: Again since this is a very short list "the extra memory a list has over a tuple" is very small. List is easier to edit otherwise both should work. I don't know why I left this comment in the first place. This will have a very minimal effect anyways so do as you like. My bad!

ladybug/legend.py
mostaphaRoudsari: I think your work can be considered as being in "debug mode" but if you think it is useful let's keep it. I think `__repr__` and `ToString` be identical as they are technically the same.

ladybug/resultmesh.py
chriswmackey: I have been going back and forth about whether this object should be specific to displaying colored meshes of studies (ie. daylight, thermal comfort, color zones) or whether it should be generalized for all geometry so we can use it for sunpath, wind rose, hourly chart, etc.  You are right that all that we would have to do for the latter is to have people input a min and max point instead of a mesh (I think just min_point and max_point is better than a special bounding_box object because specifying the points allows more flexibility).  And I guess the only extra code people would need to add outside of the general object to get it to work for colored meshes is pass the `mesh.min`, `mesh.max`, and write `mesh.colors = legend.value_colors`.  Ok. I will re-make it so that it's a bit more general.

ladybug/legend.py
mostaphaRoudsari: If it was up to me I would consider most of the inputs here secondary and made `__init__` like this:

```python
   def __init__(self, domain=None, segments_count=None, colors=None, title=None, base_plane=None):
```

Once the legend is created one can change the rest of the properties.

For grouping you can use
- domain (min, max) instead of min, max
- legend segment_size (x, y) instead of segment_height and segment_width
- font and text heights can also be grouped as text properties.

ladybug/color.py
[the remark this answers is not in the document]
chriswmackey: Very fair point.  I had kinda felt the `__setitem__` here was a bit forced.  I will take it out.

ladybug/legend.py
chriswmackey: Ok. Glad to hear it. We can always change it if we realize it is too unwieldy.

ladybug/legend.py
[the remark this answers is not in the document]
chriswmackey: Ok. I will take this out.

ladybug/legend.py
chriswmackey: Ok. I will implement the `__init__` like you suggest but I don't agree that grouping min and max together here is a good idea.  If they are grouped, there is no way to tell if only the min is default and not the max and vice versa.  This becomes detrimental as you get to the other objects that use LegendParameters.  The segment_size and text_height suffers from a similar situation since the default values for these are all inter-related (allowing the legend to scale as a whole by setting the segment_height).  Also grouping only two items together at once doesn't seem to change things much.  If we were to group 3 or more things together (maybe base plane, all of the interrelated widths/heights), that might make a difference.  However, doing this seems a bit arbitrary.

chriswmackey: Done.  I removed the ability to set many of the parameters with `__init__`  and they are now only set-able with the setters.

chriswmackey:
The last commit I pushed should address all of your comments, @mostaphaRoudsari .

ladybug/legend.py
mostaphaRoudsari: Looks good. thanks.

chriswmackey:
Awesome. Thanks @mostaphaRoudsari .
Merged!

PR #146, opened 2019-06-08
refactor(dict): Replacing to_json and from_json with to_dict and from_dict

chriswmackey:
... there are also a few minor edits to the psychrometric library  but these do not require review as it's mostly adding docstrings and tests.


chriswmackey:
I'm going to close this out as there is another PR that I wanted to put into the next release.

PR #150, opened 2019-06-21
fix(pypi): fix deployment issue to pypi

### LINKED ISSUE #149: PyPi deployment failed
The latest PyPi deployment failed because of something in the long description section not rendering properly. Not sure how or why as it's worked in the past.

https://travis-ci.org/ladybug-tools/ladybug/jobs/546918735#L918

mostaphaRoudsari:
I updated the libraries locally and was able to create the package successfully. I'm sending the PR directly to master so we see if it is really fixing the issue.

also see: https://github.com/pypa/warehouse/issues/4079#issuecomment-420929642

fixes #149

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ ladybug/sunpathplus.py
 *.ipynb
 .tox
 *.egg-info
+.eggs/*
 build
 dist
 venv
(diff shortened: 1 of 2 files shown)


mostaphaRoudsari:
I merged this in based on @AntoineDao's 👍

PR #154, opened 2019-08-15
feat(datatype): Adding data types for schedule type limits

chriswmackey:
Now that I am starting to implement schedule in honeybee-energy, I realized there are a few more data types that it would be useful to have in ladybug core.

### DIFF
--- a/ladybug/datatype/power.py
+++ b/ladybug/datatype/power.py
@@ -69,3 +69,8 @@ class Power(DataTypeBase):
     def isPower(self):
         """Return True."""
         return True
+
+
+class ActivityLevel(Power):
+    _abbreviation = 'Activity'
+    _min = 0
--- a/ladybug/datatype/uvalue.py
+++ b/ladybug/datatype/uvalue.py
@@ -41,3 +41,11 @@ class UValue(DataTypeBase):
     def isUValue(self):
         """Return True."""
         return True
+
+
+class ConvectionCoefficient(UValue):
+    _abbreviation = 'hc'
+
+
+class RadiantCoefficient(UValue):
+    _abbreviation = 'hr'


chriswmackey:
I was originally using the imperative tense but then I thought it looked weird in the auto-generated release notes to not report what "was added" or what "we are adding" but instead report what "we add". I mean, I'll follow the standard if it's the standard but I'm having difficulty understanding where this one came from.

mostaphaRoudsari:
image

AntoineDao:
> LGTM!
> 
> On behalf of @AntoineDao I have to add that the commit message should be an "imperative tense description of the change". 🚨git🚨 😄

My name is Antoine Dao and I partially approve this message. Main thing I'm fussed about is getting overall formatting so it doesn't break the release process. Actual tense of the message becomes personal preference, guess it will mostly look weird if we don't use the same tense (but I don't care enough or have the energy to debate this at length 😛 )

mostaphaRoudsari:
I think being consistent is the most important and imperative tense actually looks good to me. At the same time it falls under [bike-shed color problem](http://bikeshed.org/) category. As a result I happily follow what has been recommended by `commitizen-cli`.

chriswmackey:
Thank you both and I did a little research myself after I merged this in.  The only real explanation that I could find across all of the forums and committing guidelines that I searched is this:  That git uses the imperative tense when it auto-generates commit headers for you.  For example:
"Merged development into master"
"Update README.md".  
So, for the sake of consistency, all developers are encouraged to use the imperative mood in their commit headers.  I honestly don't care enough to debate it either since this is a case where it's more important that a standard exists than that the standard has the fewest negative consequences.  Still, it is just easier for me to remember to follow the standard when I have a reason like "it's what git does".

AntoineDao:
But... `we've always done it this way...` :: 👅

mostaphaRoudsari:
Now and since we have everything else figured out let's discuss if we should use all lowercase or capitalize the first letter! 🆎 / 🔤 =))

Issue #129, opened 2019-03-29
Ensure documentation is not overwritten by empty


AntoineDao:
As per https://github.com/ladybug-tools/ladybug/pull/128 we are having issues with empty docs being generated.

mostaphaRoudsari:
This is addressed by @AntoineDao!

PR #155, opened 2019-08-17
ci(docs): build docs in seperate step to PyPi deployment

AntoineDao:
#129

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,1 @@
+ladybug-geometry==1.6.2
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -0,0 +1 @@
+ladybug-geometry==1.6.2
```
mostaphaRoudsari: I think this makes sense to keep the requirements to a released version rather than the development branch but I think @chriswmackey is using the development branch here for a reason.

PR #156, opened 2019-08-17
feat(date, time): Add objects for Date and Time

chriswmackey:
These two objects are particularly helpful for defining EnergyPlus schedules and there may be some other places where it is helpful to have these objects coordinated with the Ladybug DateTime and AnalysisPeriod as they are here.

This commit also adds a full set of tests for DateTime, Date, and Time.  It also includes a few corrections to places where we were internally using hasattr instead of isinstance.

### DIFF
--- a/ladybug/dt.py
+++ b/ladybug/dt.py
@@ -2,19 +2,23 @@
 """Ladybug datetime."""
 from __future__ import division
 
-from datetime import datetime
+from datetime import datetime, date, time
 
 
 class DateTime(datetime):
     """Create Ladybug Date time.
 
-    Args:
-        month: A value for month between 1-12 (Defualt: 1).
-        day: A value for day between 1-31 (Defualt: 1).
-        hour: A value for hour between 0-23 (Defualt: 0).
-        minute: A value for month between 0-59 (Defualt: 0).
-        leap_year: A boolean to indicate if datetime is for a leap year
-            (Default: False).
+    Properties:
+        month
+        day
+        hour
+        minute
+        leap_year
+        doy
+        hoy
+        moy
+        int_hoy
+        float_hour
     """
 
     __slots__ = ()
@@ -31,7 +35,7 @@ class DateTime(datetime):
                 (Default: False).
         """
         year = 2016 if leap_year else 2017
-        hour, minute = cls._calculate_hour_and_minute(hour + minute / 60.0)
+        hour, minute = Time._calculate_hour_and_minute(hour + minute / 60.0)
         try:
             return datetime.__new__(cls, year, month, day, hour, minute)
         except ValueError as e:
@@ -51,23 +55,12 @@ class DateTime(datetime):
                 'minute': A value for month between 0-59. (Defualt: 0)
             }
         """
-        if 'month' not in data:
-            data['month'] = 1
-
-        if 'day' not in data:
-            data['day'] = 1
-
-        if 'hour' not in data:
-            data['hour'] = 0
-
-        if 'minute' not in data:
-            data['minute'] = 0
-
-        if 'year' not in data:
-            data['year'] = 2017
-
-        leap_year = True if int(data['year']) == 2016 else False
-        return cls(data['month'], data['day'], data['hour'], data['minute'], leap_year)
+        month = data['month'] if 'month' in data else 1
+        day = data['day'] if 'day' in data else 1
+        hour = data['hour'] if 'hour' in data else 0
+        minute = data['minute'] if 'minute' in data else 0
+        leap_year = data['leap_year'] if 'leap_year' in data else False
+        return cls(month, day, hour, minute, leap_year)
 
     @classmethod
     def from_hoy(cls, hoy, leap_year=False):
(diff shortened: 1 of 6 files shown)


ladybug/dt.py
```diff
@@ -220,3 +224,199 @@ def ToString(self):
...
+            return date.__new__(cls, year, month, day)
+        except ValueError as e:
+            raise ValueError("{}:\n\t({}/{})(m/d)".format(e, month, day))
+
+    @classmethod
+    def from_dict(cls, data):
+        """Create date from a dictionary.
+
+        Args:
+            data: {
```
mostaphaRoudsari: This will not render correctly in documentation. I remember that we fixed it once but don't remember for which library

You should use something like

```
.. code-block:: python

  data = { this: that}
```
see here: https://www.sphinx-doc.org/en/1.5/markup/code.html#directive-code-block

ladybug/dt.py
```diff
@@ -220,3 +224,199 @@ def ToString(self):
...
+        """
+        if not leap_year:
+            days_until_month = (0, 31, 59, 90, 120, 151, 181,
+                                212, 243, 273, 304, 334, 365)
+        else:
+            days_until_month = (0, 31, 60, 91, 121, 152, 182,
+                                213, 244, 274, 305, 335, 366)
+
+        for month_count in range(12):  # find month
+            if int(doy) < days_until_month[month_count + 1]:
```
mostaphaRoudsari: I would take `int(doy)` out of the loop. That will also ensure that your code in line 209 doesn't break. for illegal values you can use for / else pattern.

```python
doy = int(doy)
for month_count in range(12):  # find month
    if doy < days_until_month[month_count + 1]:
        month = month_count + 1
        break
else:
    raise ValueError(
        "doy must be positive and smaller than 366. Invalid input %d" % (doy)
    )

day = int(doy - days_until_month[month - 1])
```

ladybug/dt.py
```diff
@@ -220,3 +224,199 @@ def ToString(self):
...
+        return self.year == 2016
+
+    @property
+    def doy(self):
+        """Calculate day of the year for this date."""
+        return self.timetuple().tm_yday
+
+    def to_dict(self):
+        """Get date as a dictionary."""
+        base = {'month': self.month, 'day': self.day}
```
mostaphaRoudsari: are we going to include `type` key?

ladybug/dt.py
```diff
@@ -220,3 +224,199 @@ def ToString(self):
...
+    __slots__ = ()
+
+    def __new__(cls, hour=0, minute=0):
+        """Create Ladybug Time.
+
+        Args:
+            hour: A value for hour between 0-23 (Defualt: 0).
+            minute: A value for month between 0-59 (Defualt: 0).
+        """
+        hour, minute = DateTime._calculate_hour_and_minute(hour + minute / 60.0)
```
mostaphaRoudsari: Probably it is more natural to have this method in Time class and reference it in DateTime not that other way around.

ladybug/dt.py
```diff
@@ -220,3 +224,199 @@ def ToString(self):
...
+        return self.hour * 60 + self.minute
+
+    @property
+    def float_hour(self):
+        """Get hour and minute as a float value (e.g. 6.25 for 6:15)."""
+        return self.hour + self.minute / 60.0
+
+    def to_dict(self):
+        """Get time as a dictionary."""
+        return {'hour': self.hour, 'minute': self.minute}
```
mostaphaRoudsari: `type` key is missing.

ladybug/dt.py
chriswmackey: So there are a number of changes that need to be made across all of our documentation and I have put them in a list here:
https://docs.google.com/document/d/1MmKllWNW5B4TvYlDawzQxS_kq5AFXmnWIBf68uCBCA0/edit?usp=sharing

These changes need to happen across nearly all of the main repos and I think it is going to be easier to update an entire repo altogether rather than having half of it updated and the other half not updated.  So I am going to leave this as it is now and someone (possibly me in a couple of months) will go through and update all of the documentation together at some point.

ladybug/dt.py
chriswmackey: Good idea.  I will do the same thing for your old hoy code.

ladybug/dt.py
mostaphaRoudsari: Sounds like a plan! I can also add to the list. If you run sphix commands locally you will see a long list of warnings that should be addressed.

ladybug/dt.py
chriswmackey: I am thinking that we should but this is an issue much like the documentation one that you pointed out.  Currently, the `type` key isn't used across all of ladybug-core right now and I think it will be better to update it altogether.  So, again, I will leave this as it is now so that we don't get confused about whether we fixed this repo or not.  I have also opened an issue so that we don't forget about it:
https://github.com/ladybug-tools/ladybug/issues/157

ladybug/dt.py
chriswmackey: Also, I put together an issue so we don't forget:
https://github.com/ladybug-tools/ladybug/issues/158

ladybug/dt.py
chriswmackey: Good point and I agree

ladybug/dt.py
chriswmackey: See above.

PR #161, opened 2019-08-20
feat(datetime): Add to_array and from_array methods to datetime

chriswmackey:
This makes the dictionary representation of objects like DataCollections a little lighter.  It will also make the honyebee-energy Schedule dictionaries a little lighter.


chriswmackey:
... I am also taking the opportunity to fix some incorrect indentations and areas where I forgot to use a literal string for regexes.

Issue #158, opened 2019-08-19
Update several things in the formatting of the documentation


chriswmackey:
Several changes need to be made to the documentation of this repo:
(check means we know how to fix them)

* Put any ASCII art in a .. code-block:: shell
* Put any dictionaries in from_dict methods in a .. code-block:: json
* Move all descriptions in `__init__` to be on the class itself.
* Get child classes to display documentation of their children
* Get the returns to display correctly (not bolded). Is do-able with @santiagogaray 's [solution here](https://github.com/ladybug-tools/ladybug-geometry/blob/master/docs/conf.py#L233-L285)
* Get properties and methods to be organized groupwise
* Get the navigation bar to not bleed onto the docs. Is doable with @santiagogaray 's solution [here](https://github.com/ladybug-tools/ladybug-geometry/blob/master/docs/_templates/layout.html), [here](https://github.com/ladybug-tools/ladybug-geometry/blob/master/docs/_static/custom.css)  and [here](https://github.com/ladybug-tools/ladybug-geometry/blob/master/docs/conf.py#L288-L298).
* Get all `Args` docstrings to use a standard on how to include the default values. Specifically, this should be `Default: False` at the end of the argument description (note that `False` should be replaces with the actual default value).

If anyone is up to the task, please go for it.  Otherwise, I can get to is after a few months.

Kiy4h:
May I work on this? I'll be asking questions though.

santiagogaray:
I actually have been working on it this week, I'm almost done with it.

Kiy4h:
Oh whoops, sorry! I'll try to look for something else :)

santiagogaray:
I'm sorry, my bad for not posting about it here!

chriswmackey:
Sorry I could have been clearer about assigning this issue to @santiagogaray . @Kiy4h , if you want to see everything that we need help on, check our [hacktoberfest board](https://github.com/orgs/ladybug-tools/projects/2)

chriswmackey:
Thanks again for addressing this one @santiagogaray .  This was a huge one!

PR #170, opened 2019-10-07
Update several things in the formatting of the documentation #158

santiagogaray:
@chriswmackey @mostaphaRoudsari @theo-armour 
Took a while but I finally completed a first pass on the entire package. I fixed many display/formatting issues but a few things remain. If you have time read on, help is appreciated.

1a)
to_dict() method:
Throughout the package, in many from_dict methods, the docstring shows the dictionary with the description of the keys, this makes sphinx to invalidate it as  json code. 
Could possibly solve it in a number of different ways, just a couple :
a) Not using json block and use a literal or shell block.
b) As shown in ladybug.analysisperiod.AnalysisPeriod: writing the dictionary with default values only and listing the keys outside of the dictionary.

1b) 
Many from_dict() cases fail to be detected as json given the differences between dictionaries and json formats:
Python uses single and double quotes for string literals. JSON only allows double quotes.
Tuples not supported in JSON
Python uses None objects, JSON uses null to as a special "doesn't exist" sentinel value.
Python uses True and False as boolean values, JSON uses true and false.
Python dictionary keys can be any hashable type, JSON only supports strings.
Do we JSONize the python dictionaries or we specify the block as something else(ex. code).

2)
In many classes I added properties when missing. In some classes  Properties / Arguments that listed _init_ inputs were renamed as Args before adding the properties. 
Do we add descriptions to the properties?  We have inconsistency at this point, sometimes in Arguments we have descriptions but not in properties (sometimes are the same, more often than not they differ)

3)
Usage - Is correct to use a literal block for these? Should title be bolded?

4)
Are constants displayed correctly? Right now contents are shown bolded, in some modules they extend multiple lines (ex. ladybug.epw.EPW.FIELDS)

6) 
In Class properties where descriptions extend multiple lines (ex. ladybug.analysisperiod.AnalysisPeriod)  lines 2 and beyond display incorrectly once the property name is made bolded. We actually have this issue in one instance in ladybug.geometry(geometry3d.arc for reference). After trying to solve this in the docstring itself, the only way I seem to find a workaround was by joining these multi-line descriptions in one single line. This was done at the  autodoc_process_docstring by adding a new function, make_properties_one_liner, to execute before make_fields_bolded. I may find a way to simplify this routine but seems to be doing the job well at this point. 
If you think there is a simpler way to tackle this issue feel free to pitch in, any comments is appreciated.
 
7) 
Minor observations - not documentation related:
ladybug.datacollection.HourlyContinuousCollection: Has is_continuous and isContinuous properties

VolumeFlowRate:
Shouldn't isFlowRate property be isVolumeFlowRate?

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -223,7 +223,7 @@ autodoc_member_order = 'groupwise'
 
 
 def setup(app):
-    """Run custom code with access to the Shinx application object
+    """Run custom code with access to the Sphinx application object
 
     Args:
         app: the Sphinx application object
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,8 +6,6 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
-import re
-
 # -- Path setup --------------------------------------------------------------
 
 # If extensions (or modules to document with autodoc) are in another directory,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -6,6 +6,8 @@
 # full list see the documentation:
 # http://www.sphinx-doc.org/en/master/config
 
+import re
+
 # -- Path setup --------------------------------------------------------------
 
 # If extensions (or modules to document with autodoc) are in another directory,
@@ -100,11 +102,11 @@ html_theme_options = {
     # Fix navigation bar to top of page?
     # Values: "true" (default) or "false"
     'navbar_fixed_top': "true",
-	'navbar_pagenav': True,
+    'navbar_pagenav': True,
     'source_link_position': "nav",
-	'bootswatch_theme': "united",
+    'bootswatch_theme': "united",
     'bootstrap_version': "3",
-	}
+}
 
 # on_rtd is whether we are on readthedocs.org
 # on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
@@ -213,3 +215,119 @@ epub_exclude_files = ['search.html']
 
 # If true, `todo` and `todoList` produce output, else they produce nothing.
 todo_include_todos = True
+
+# -- Options for autodoc extension --------------------------------------------
+autodoc_default_options = {
+    'inherited-members': True,
+}
+
+autodoc_member_order = 'groupwise'
+
+
+# Autodoc event handlers
+def autodoc_process_docstring(app, what, name, obj, options, lines):
+    """Function to handle the autodoc-process-docstring event.
+
+    Emitted when autodoc has read and processed a docstring. lines is a list of
+    strings – the lines of the processed docstring – that the event handler
+    can modify in place to change what Sphinx puts into the output.
+
+    More information on sphinx documentation:
+        http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
+        #event-autodoc-process-docstring
+
+    Args:
+        app: the Sphinx application object
+        what: the type of the object which the docstring belongs to (one of "module",
+            "class", "exception", "function", "method", "attribute")
+        name: the fully qualified name of the object
+        obj: the object itself
+        options: the options given to the directive: an object with attributes
+            inherited_members, undoc_members, show_inheritance and noindex that
+            are true if the flag option of same name was given to the auto directive
+        lines: the lines of the docstring
+    """
+
+    # Make class properties display matching the 'Args' format
+    if what == 'class':
+        make_properties_one_liner(lines)
+        make_fields_bolded(lines)
+
+    return lines
+
+
+def make_properties_one_liner(lines):
+    """Ensures that lines with the 'field: value' format that extend multiple
+    lines are merged into a single line. The multiline properties display secondary
+    lines incorrectly when field is made bolded.
+
+    Args:
+        lines: the lines of the docstring
+    """
+
+    # Connect the list strings into a single string
+    doc_text = "\n".join(lines) + "\n"
+
+    # Generate groups based on three possible line formats:
+    # 1: new line (\n)
+    # 2: Bulleted field: value (?:(\s*\*\s*)(\w+)(\s*:)([^\n]+)\n)
+    # 3: Any other string: (?:(\s*)([^\n]+)\n)
+    matches = re.findall(
+        r"(\n)|(?:(\s*\*\s*)(\w+)(\s*:)([^\n]+)\n)|(?:(\s*)([^\n]+)\n)", doc_text)
+
+    # Traverse groups to detect and join multilined field:value cases
+    i, field_i = 0, -1
+    while i < len(matches):
+        group = matches[i]
+        # Check for bulleted field:value line and track line index
+        if "*" in group[1]:
+            field_i = i
+        # Check if it is next line of field:value line(not bulleted & text vertically
+        # aligned)
+        # Connect this line's text group to parent field:value line
+        elif group[6] != '' and field_i >= 0 and group[6].lstrip()[0] != "*" and \
+                len(group[5]) == len(matches[field_i][1]):
+            lines[field_i] = lines[field_i] + ' ' + group[6]
+            lines[i] = ""
+        # If not aligned text line or empty line reset field:value line index
+        else:
+            field_i = -1
+        # Loop through every line match
+        i += 1
+
+
+def make_fields_bolded(lines):
+    """Make field font bolded in 'field: value' line format.
+
+    Note the purpose of this change is to match the class 'Parameters'
+    and 'Args' fields format.
+    Regex pattern "(\s*\*\s*)(\w+)(\s*:)(.+)" detects range of 'field: value'
+    formats.
+
+    Args:
+        lines: the lines of the docstring
+    """
+
+    # Include boldface (reST inline markup '**') in field names of matching lines
+    def replace(match):
+        return "{0[0]} **{0[1]}** --{0[3]}".format(match.groups())
+
+    # Substitute macthing lines
+    new_lines = [re.sub(r"(\s*\*\s*)(\w+)(\s*:)(.+)", replace, line) for line in lines]
+
+    # Copy changes back to lines (in place)
+    for i in range(len(lines)):
+        lines[i] = new_lines[i]
+
+
+def setup(app):
+    """Run custom code with access to the Shinx application object
+
+    Args:
+        app: the Sphinx application object
+    """
+    # Register the event handler for 'autodoc-process-docstring' event
+    app.connect('autodoc-process-docstring', autodoc_process_docstring)
+
+    # Add bootstrap theme custom stylesheet
+    app.add_stylesheet("custom.css")
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -224,110 +224,12 @@ autodoc_default_options = {
 autodoc_member_order = 'groupwise'
 
 
-# Autodoc event handlers
-def autodoc_process_docstring(app, what, name, obj, options, lines):
-    """Function to handle the autodoc-process-docstring event.
-
-    Emitted when autodoc has read and processed a docstring. lines is a list of
-    strings – the lines of the processed docstring – that the event handler
-    can modify in place to change what Sphinx puts into the output.
-
-    More information on sphinx documentation:
-        http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html
-        #event-autodoc-process-docstring
-
-    Args:
-        app: the Sphinx application object
-        what: the type of the object which the docstring belongs to (one of "module",
-            "class", "exception", "function", "method", "attribute")
-        name: the fully qualified name of the object
-        obj: the object itself
-        options: the options given to the directive: an object with attributes
-            inherited_members, undoc_members, show_inheritance and noindex that
-            are true if the flag option of same name was given to the auto directive
-        lines: the lines of the docstring
-    """
-
-    # Make class properties display matching the 'Args' format
-    if what == 'class':
-        make_properties_one_liner(lines)
-        make_fields_bolded(lines)
-
-    return lines
-
-
-def make_properties_one_liner(lines):
-    """Ensures that lines with the 'field: value' format that extend multiple
-    lines are merged into a single line. The multiline properties display secondary
-    lines incorrectly when field is made bolded.
-
-    Args:
-        lines: the lines of the docstring
-    """
-
-    # Connect the list strings into a single string
-    doc_text = "\n".join(lines) + "\n"
-
-    # Generate groups based on three possible line formats:
-    # 1: new line (\n)
-    # 2: Bulleted field: value (?:(\s*\*\s*)(\w+)(\s*:)([^\n]+)\n)
-    # 3: Any other string: (?:(\s*)([^\n]+)\n)
-    matches = re.findall(
-        r"(\n)|(?:(\s*\*\s*)(\w+)(\s*:)([^\n]+)\n)|(?:(\s*)([^\n]+)\n)", doc_text)
-
-    # Traverse groups to detect and join multilined field:value cases
-    i, field_i = 0, -1
-    while i < len(matches):
-        group = matches[i]
-        # Check for bulleted field:value line and track line index
-        if "*" in group[1]:
-            field_i = i
-        # Check if it is next line of field:value line(not bulleted & text vertically
-        # aligned)
-        # Connect this line's text group to parent field:value line
-        elif group[6] != '' and field_i >= 0 and group[6].lstrip()[0] != "*" and \
-                len(group[5]) == len(matches[field_i][1]):
-            lines[field_i] = lines[field_i] + ' ' + group[6]
-            lines[i] = ""
-        # If not aligned text line or empty line reset field:value line index
-        else:
-            field_i = -1
-        # Loop through every line match
-        i += 1
-
-
-def make_fields_bolded(lines):
-    """Make field font bolded in 'field: value' line format.
-
-    Note the purpose of this change is to match the class 'Parameters'
-    and 'Args' fields format.
-    Regex pattern "(\s*\*\s*)(\w+)(\s*:)(.+)" detects range of 'field: value'
-    formats.
-
-    Args:
-        lines: the lines of the docstring
-    """
-
-    # Include boldface (reST inline markup '**') in field names of matching lines
-    def replace(match):
-        return "{0[0]} **{0[1]}** --{0[3]}".format(match.groups())
-
-    # Substitute macthing lines
-    new_lines = [re.sub(r"(\s*\*\s*)(\w+)(\s*:)(.+)", replace, line) for line in lines]
-
-    # Copy changes back to lines (in place)
-    for i in range(len(lines)):
-        lines[i] = new_lines[i]
-
-
 def setup(app):
     """Run custom code with access to the Shinx application object
 
     Args:
         app: the Sphinx application object
     """
-    # Register the event handler for 'autodoc-process-docstring' event
-    app.connect('autodoc-process-docstring', autodoc_process_docstring)
 
     # Add bootstrap theme custom stylesheet
     app.add_stylesheet("custom.css")
--- a/ladybug/datacollection.py
+++ b/ladybug/datacollection.py
@@ -105,11 +105,11 @@ class HourlyDiscontinuousCollection(BaseCollection):
         .. code-block:: python
 
                 {
-                "header": h  # A Ladybug Header,
-                "values": v  # An array of values,
-                "datetimes": d  # An array of datetimes,
-                "validated_a_period": p  # Boolean for whether header
-                                          # analysis_period is valid
+                "header": {}  # A Ladybug Header,
+                "values": []  # An array of values,
+                "datetimes": []  # An array of datetimes,
+                "validated_a_period": True  # Boolean for whether header
+                                            # analysis_period is valid
                 }
         """
         assert 'header' in data, 'Required keyword "header" is missing!'
@@ -603,8 +603,8 @@ class HourlyContinuousCollection(HourlyDiscontinuousCollection):
         .. code-block:: python
 
                 {
-                "header": h  # A Ladybug Header,
-                "values": v  # An array of values,
+                "header": {}  # A Ladybug Header,
+                "values": []  # An array of values,
                 }
         """
         assert 'header' in data, 'Required keyword "header" is missing!'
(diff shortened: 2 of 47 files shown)


santiagogaray:
from_dict()  possible solutions

image
image

santiagogaray:
Usage display

image

santiagogaray:
Constants display

image

santiagogaray:
Multiline property issue. We have this one case in ladybug-geometry. 

image

Fixed in this package

chriswmackey:
Thank you very much for getting this together @santiagogaray . I will try to review it as soon as I can.

chriswmackey:
Thank you again for getting this together @santiagogaray . I am going to answer your high-level questions first and then do a review looking through all of your changes.

1. I think we want to standardize the display of dictionaries in docstrings and I think the convention we used in `ladybug-geometry` is better. So, if you can add a sample value for each key in the docstring dictionary (like a float, int, empty list, empty dictionary) and then put the description behind a comment like `\\` or `#`, that would be ideal.  BTW, I am thinking that we should probably be using `codeblock::python` instead of `codeblock::json`. The input dictionary for this method is a Python one after all, even though it is easily convert-able to a JSON format. So using `codeblock::python` will solve all of the issues you experience there about the difference between Python and JSON.

2. This repo is the oldest of all the Ladybug Tools organization repos and so it has a lot of docstring stuff that is not following the standards we are trying to follow today.  The `ladybug-geometry` repo is much newer and so I would generally say that, when in doubt, do what we did on that repo. For clarification, this means: 1. Adding a bulleted list of properties (without descriptions) at the top of each class. 2. Using an `Args:` tag at the top of the class that describes each input going into the `__init__` method. 3. Including the descriptions of each property on the docstring of that property (not within the bulleted list at the top).

3. Usage should always be in `codeblock::python`.

4. The `FIELDS` constant really is not that helpful and I think we should make it hidden because, otherwise, it makes the docs really unreadable. So changing `FIELDS` to `_fields` should do the trick in this situation.  For the other constants, I think we usually want them to show up in the docs (like I know I definitely want to see the 'VALIDTIMESTEPS' constant on the `AnalysisPeriod` to show up). For this situation, I think the boded text makes sense. Feel free to tweak it if you think it can be improved.

6. I think this one should be solved by following my suggestion above. Namely, let's refrain from having a description of the property within the bulleted list and instead put the description of the property within the docstring of the property itself.

7. Thank you for the observations. Actually, all of these situations were we have `isSuchAndSuch` properties is not good practice and we should remove them at some point soon. It's a part of a hack that Mostapha had done to get the library to work in the (now already broken) Dynamo version, which we realize now is not worth making the code here un-clean. We will have to remove both these `isSuchAndSuch` properties and any cases where they are used in `hasattr` methods throughout the library. We can do this all in a later PR, though, so you don't have to worry about it now. Perhaps the only bearing that this has on your docstrings is that you can leave these `isSuchAndSuch` properties out of the bulleted list of properties at the top.

docs/conf.py
```diff
@@ -213,3 +215,119 @@
...
+
+    # Make class properties display matching the 'Args' format
+    if what == 'class':
+        make_properties_one_liner(lines)
+        make_fields_bolded(lines)
+
+    return lines
+
+
+def make_properties_one_liner(lines):
```
chriswmackey: As I said in my previous answers to your questions, I think we can remove this and just use better practices of not describing inputs in the bulleted list of properties.

santiagogaray:
Thanks @chriswmackey for your comments. I have updated the package with the changes suggested. 

A couple of things to check:

-The values used in the dictionaries can be example values, empty data structures, or variables. Feel free to indicate if any of these approaches are not working.

- I removed all the property descriptions included in the DataTypeBase class. Given these descriptions are more extensive than the ones included in the property methods, I used the former descriptions to replace the existing ones. Is this amount of information correct to be included here?

- As you can see the conf.py got a lot simpler now, without any property descriptions there is no need for any post-processing tasks . I'm assuming this will have to be updated in the ladybug_geometry package to be consistent, right?

Thanks again Chris, look forward to your comments.

chriswmackey:
@santiagogaray ,

Thanks for addressing the comments and I think we still have one more round of me answering high-level questions before I dig in deep.

Firs and foremost, you will see that one of your recent changes has broken the code, causing a particularly high number of failures in the `EPW` and `Wea` classes that you can see here: https://travis-ci.org/ladybug-tools/ladybug/jobs/596200537
I think this might have resulted from you making changes from `FIELDS` to `_fields` without changing all of the places where the original `FIELDS` constant is referenced.  We will need to get this fixed before I can do a deep review and you can run the tests locally from command line by using the following steps:
1. Install pytest by typing `pip install pytest`
2. Navigate your command window to the root of the reop on your local machine.
3. Type `pytest .\tests`
Then you should see all the tests will run and you'll get info about all of the errors. Note that you may need to include `python -m` before each of your commands depending on how Python has been added to your PATH.

I'll answer your other questions shortly.

santiagogaray:
Thanks @chriswmackey for guiding me with the pytest tool. After renaming a couple of FIELDS I checked again and no errors reported.  Happy to learn yet a new thing related to this process!

1. I did not explain myself well in regards to the dictionaries and sample values.  I will show with examples in a separate comment below.
 
2. I did tweak a few properties' docstrings in DataTypeBase. Hope is better, happy to simplify more if needed.

3. I had fun learning regex, thanks for giving the opportunity to learn it!

santiagogaray:
When deciding the dictionary sample values I was using variable names instead of actual values in dictionaries where some keys hold some unknown data structure (ex. header below). What would you do in this case? Is header here an empty dictionary?

image

chriswmackey:
Thank you for fixing the failing tests @santiagogaray . As you can see, the automated run of the tests with each new commit is important and I'm happy to show you more about how they work at a later point.

Ah, that's what you meant by the dictionaries. For these, what you have there is close but, for the case where you have another object nested within the parent one, the dictionary representation will have another nested dictionary. So instead of `h`, you should use some empty curly brackets `{}` to denote that the value there is another dictionary. Similarly, in the case that the object is an array of something, you should use empty square brackets `[]`.  Altogether, the dictionary that you screen-shot there should look like this:
```
                {
                "header": {}  # A Ladybug Header,
                "values": []  # An array of values,
                "datetimes": []  # An array of datetimes,
                "validated_a_period": True  # Boolean for whether analysis_period is valid
                }
```

BTW, I forgot to confirm with you that it will be good to go back to ladybug-geometry and remove the descriptions from the bulleted list of properties. Feel free to send a PR for that whenever you want and we should be able to merge it in quickly.

chriswmackey:
Also, once you fix those dictionary displays, I'll do a final review.

santiagogaray:
Thanks @chriswmackey for clarifying. I hope I got this right this time and you can start your review.

I will get going on the ladybug_geoemtry package soon.

I appreciate the patience and time to guide me through all the steps involved in this. I'm grateful I have a chance to learn as we go through things. Thanks!

ladybug/designday.py
```diff
@@ -1136,11 +1205,21 @@ def __repr__(self):
...
     """An object representing a sky on the design day.
 
-    Properties:
+    Args:
         solar_model
         month
         day_of_month
         daylight_savings_indicator
+        beam_shced
+        diff_sched
```
chriswmackey: I also forgot the docstring here. Can you add the following:
```
        solar_model: Text for the name of the solar model to use. Choose from the
            following: 'ASHRAEClearSky', 'ASHRAETau', 'ZhangHuang', 'Schedule'
        month: Month in which the design day occurs.
        day_of_month: Day of the month on which the design day occurs.
        daylight_savings_indicator: Text ('Yes' or 'No'), for whether daylight savings
            time is active. Default: 'No'
        beam_shced: Schedule name for beam irradiance. Shoulb be an empty string unless
            solar_model is 'Schedule'.
        diff_sched: Schedule name for diffuse irradiance. Shoulb be an empty string
            unless solar_model is 'Schedule'.
```

ladybug/designday.py
```diff
@@ -1012,11 +1066,21 @@ def __repr__(self):
 class WindCondition(object):
     """Represents wind and rain conditions on the design day.
 
-    Properties:
+    Args:
         wind_speed
         wind_direction
         rain
         snow_on_ground
+
```
chriswmackey: I forgot the docstring here. Can you add the following:
```
wind_speed: Wind speed on the design day [m/s].
        wind_direction: Wind direction on the design day [degrees]. Default: 0
        rain: Boolean to indicate rain on the design day. Default: False.
        snow_on_ground: Boolean to indicate snow on the design day. Default: False.
```

ladybug/designday.py
```diff
@@ -1273,11 +1357,17 @@ def __repr__(self):
 class OriginalClearSkyCondition(SkyCondition):
     """An object representing an original ASHRAE Clear Sky.
 
-    Properties:
+    Args:
         month
         day_of_month
         clearness
         daylight_savings_indicator
+
```
chriswmackey: I see I forgot the docstring here. Can you add the following:
```
        month: Month in which the design day occurs.
        day_of_month: Day of the month on which the design day occurs.
        clearness: Value between 0 and 1.2 that will get multiplied by the model's
            irradinace to correct for factors like elevation.
        daylight_savings_indicator: Text ('Yes' or 'No'), for whether daylight savings
            time is active. Default: 'No'
```

ladybug/designday.py
```diff
@@ -1369,12 +1464,22 @@ def to_dict(self):
...
     """An object representing an ASHRAE Revised Clear Sky (Tau model).
 
-    Properties:
+    Args:
         month
         day_of_month
         tau_b
         tau_d
         daylight_savings_indicator
+
```
chriswmackey: I see I forgot a docstring here. Can you add the following:
```
        month: Month in which the design day occurs.
        day_of_month: Day of the month on which the design day occurs.
        tau_b: Value for the 'beam' term in the Tau model. Typically
            found in .stat files.
        tau_d: Value for the 'diffuse' term in the Tau model. Typically
            found in .stat files.
        daylight_savings_indicator: Text ('Yes' or 'No'), for whether daylight savings
            time is active. Default: 'No'
```

ladybug/dt.py
```diff
@@ -256,23 +271,26 @@ def __repr__(self):
 class Date(date):
     """Ladybug Date.
 
+    Args:
+        month: A value for month between 1-12 (Defualt: 1).
+        day: A value for day between 1-31 (Defualt: 1).
+        leap_year: A boolean to indicate if date is for a leap year
+            (Default: False).
```
chriswmackey: Let's remove the parentheses around the Default and just add a period after `year`. This is the new standard we are trying to follow throughout the libraries.

ladybug/stat.py
```diff
@@ -120,33 +123,37 @@ def __init__(self, file_path):
...
-            }
+        Args:
+            data: A python dictionary in the following format
+
+        .. code-block:: python
+
+                {
+                "location": {},  # ladybug location schema
+                "ashrae_climate_zone": "",  # str
+                "koppen_climate_zone": "", # str
```
chriswmackey: We can put some better defaults here. How about `"5A"` for `ashrae_climate_zone` and `"Dfa"` for `koppen_climate_zone`. Those are the values for Boston.

ladybug/stat.py
```diff
@@ -120,33 +123,37 @@ def __init__(self, file_path):
...
+                "monthly_db_100": [],  # list of 12 float values for each month
+                "monthly_wb_100": [],  # list of 12 float values for each month
+                "monthly_db_20": [],  # list of 12 float values for each month
+                "monthly_wb_20": [],  # list of 12 float values for each month
+                "monthly_db_04": [],  # list of 12 float values for each month
+                "monthly_wb_04": [],  # list of 12 float values for each month
+                "monthly_wind": [],  # list of 12 float values for each month
+                "monthly_wind_dirs": [],  # matrix with 12 cols for months of the year
+                                          #and 8 rows for the cardinal directions.
+                "standard_pressure_at_elev": 0.0,  # float value for pressure in Pa
```
chriswmackey: A better default here is `101325`, which is air pressure at sea level in pascals. BTW, generally excellent job formatting these dictionaries.

ladybug/datacollection.py
```diff
@@ -2,23 +2,24 @@
...
 
 All Data Collections in this module have the ability to:
-    * max, min, bounds, average, median, total, get_percentile, get_highest/lowest values
-    * perform unit conversions on the data: to_unit, to_ip, to_si
-    * filter based on conditional statements
-    * filter based on analysis period
+* max, min, bounds, average, median, total, get_percentile, get_highest/lowest values
+* perform unit conversions on the data: to_unit, to_ip, to_si
+* filter based on conditional statements
+* filter based on analysis period
```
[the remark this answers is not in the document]
santiagogaray: Just needed to be tagged to right. Done!

ladybug/designday.py
santiagogaray: Done!

ladybug/designday.py
santiagogaray: Done!

santiagogaray:
Thanks @chriswmackey for putting time to make this detailed review. I hope I fixed all the comments, but please feel free to check if anything else needs to be done. 

I agree the documentation looks great. Seeing Sphinx and bootstrap working together by capturing the docstrings is really special. I'm very excited that I've been able to contribute and learn though this project. 

Tomorrow I will go back to the ladybug-package and make those adjustments  as we discussed!

chriswmackey:
Great! Thank you @santiagogaray ! I will do a final build of the docs once I get home and, if all looks good, we'll merge it in.

ladybug/designday.py
```diff
@@ -31,9 +31,27 @@
...
+        * hourly_sky_cover
+        * hourly_solar_radiation
+        * hourly_wind_direction
+        * hourly_wind_speed
+        * humidity_condition
+        * isDesignDay
+        * location
+        * name
+        * sky_condition
+        * wind_condition
```
chriswmackey: I just realized that these properties are in the incorrect location. They shouldn't be non the `DDY` class and they should instead be on the `DesignDay` class.  The only properties that should appear here on the `DDY` object are:
```
* file_path
* location
* design_days
```

ladybug/designday.py
```diff
@@ -222,6 +245,15 @@ class DesignDay(object):
...
         sky_condition: Ladybug SkyCondition object
+
+    Properties:
+        * name
+        * day_type
+        * location
+        * dry_bulb_condition
+        * humidity_condition
+        * wind_condition
+        * sky_condition
```
chriswmackey: As noted above, this list should be extended with the properties currently listed in the `DDY` class docstring.

ladybug/designday.py
```diff
@@ -866,13 +911,23 @@ def __repr__(self):
...
+
+    Properties:
+        * hum_type
+        * hum_value
+        * barometric_pressure
+        * schedule
+        * wet_bulb_range
+        * hourly_pressure
+        * humid_types
+        * isHumidityCondition
```
chriswmackey: Remove `isHumidityCondition` from this list since the whole property is going to be removed soon.

ladybug/designday.py
```diff
@@ -1012,11 +1072,21 @@ def __repr__(self):
...
-        wind_direction
-        rain
-        snow_on_ground
+        * wind_speed
+        * wind_direction
+        * rain
+        * snow_on_ground
+        * hourly_values
+        * hourly_wind_dirs
+        * isWindCondition
```
chriswmackey: Remove `isWindCondition`

ladybug/designday.py
```diff
@@ -1136,11 +1211,25 @@ def __repr__(self):
...
+
     Properties:
-        solar_model
-        month
-        day_of_month
-        daylight_savings_indicator
+        * month
+        * day_of_month
+        * hourly_sky_cover
+        * isSkyCondition
```
chriswmackey: Remove `isSkyCondition`

ladybug/designday.py
```diff
@@ -1369,12 +1476,25 @@ def to_dict(self):
...
+
     Properties:
-        month
-        day_of_month
-        tau_b
-        tau_d
-        daylight_savings_indicator
+        * day_of_month
+        * hourly_sky_cover
+        * isSkyCondition
```
chriswmackey: Remove `isSkyCondition`

ladybug/dt.py
```diff
@@ -8,31 +8,39 @@
...
-        float_hour
+        * month
+        * day
+        * hour
+        * leap_year
+        * doy
+        * hoy
+        * int_hoy
+        * max
+        * microsecond
```
chriswmackey: I know that, because the ladybug DateTime inherits from the native Python datetime , this `microsecond` property exists on the class. However, it is not one that is actually usable in any Ladybug workflows and, if users set this to be something other than 0, it can actually mess up certain workflows. So we would prefer to not list it here in the properties, even though I know we can't easily stop it from appearing in the docs. The same thing can be about the following properties, which should be removed from this list (as well as the `Date` and `Time` property lists):
```
* max
* min
* second
* resolution
```
Everything else can stay.

ladybug/header.py
```diff
@@ -14,26 +14,26 @@ class Header(object):
...
         analysis_period: A Ladybug analysis period (Defualt: None)
         metadata: Optional dictionary of additional metadata,
             containing information such as 'source', 'city', or 'zone'.
+
+    Properties:
+        * data_type
+        * unit
+        * analysis_period
+        * metadata
+        * isHeader
```
chriswmackey: Remove `isHeader`

ladybug/datacollection.py
```diff
@@ -1280,20 +1380,38 @@ def __repr__(self):
...
+            and the second is the hour. (eg. (12, 23) = December at 11 PM)
+
+    Properties:
+        * average
+        * bounds
+        * datetimes
+        * header
+        * isDataCollection
+        * isMonthlyPerHour
+        * isHourly
```
chriswmackey: Remove the following:
```
        * isDataCollection
        * isMonthlyPerHour
        * isHourly
```

ladybug/datacollection.py
```diff
@@ -1148,20 +1227,38 @@ def __repr__(self):
...
+            year for each value of the collection.
+
+    Properties:
+        * average
+        * bounds
+        * datetimes
+        * header
+        * isDataCollection
+        * isMonthlyPerHour
+        * isHourly
```
chriswmackey: Remove the following:
```
        * isDataCollection
        * isMonthlyPerHour
        * isHourly
```

ladybug/datacollection.py
```diff
@@ -935,20 +993,37 @@ def __repr__(self):
...
```
chriswmackey: Remove the following:
```
        * isDataCollection
        * isMonthlyPerHour
        * isHourly
```

ladybug/datacollection.py
```diff
@@ -509,19 +543,39 @@ def __repr__(self):
...
+            must align with the AnalysisPeriod on the header.
+
+    Properties:
+        * average
+        * bounds
+        * datetimes
+        * header
+        * isContinuous
+        * isDataCollection
+        * isHourly
```
chriswmackey: Remove the following:
```
        * isDataCollection
        * isMonthlyPerHour
        * isHourly
```

ladybug/datacollection.py
```diff
@@ -47,19 +48,39 @@
...
+            the list of values.
+
+
+    Properties:
+        * average
+        * bounds
+        * datetimes
+        * header
+        * isDataCollection
+        * isHourly
```
chriswmackey: Remove the following:
```
        * isDataCollection
        * isHourly
```

ladybug/designday.py
```diff
@@ -31,9 +31,27 @@
...
+        * hourly_dew_point
+        * hourly_dry_bulb
+        * hourly_horizontal_infrared
+        * hourly_relative_humidity
+        * hourly_sky_cover
+        * hourly_solar_radiation
+        * hourly_wind_direction
+        * hourly_wind_speed
+        * humidity_condition
+        * isDesignDay
```
chriswmackey: Remove `isDesignDay`

ladybug/designday.py
```diff
@@ -764,11 +791,24 @@ def __repr__(self):
...
+
     Properties:
-        dry_bulb_max
-        dry_bulb_range
-        modifier_type
-        modifier_schedule
+        * dry_bulb_max
+        * dry_bulb_range
+        * hourly_values
+        * isDryBulbCondition
```
chriswmackey: Remove `isDryBulbCondition`

ladybug/location.py
```diff
@@ -18,6 +18,20 @@ class Location(object):
         elevation: A number for elevation of the location.
         station_id: ID of the location if the location is represnting a weather station.
         source: Source of data (e.g. TMY, TMY3).
+
+    Properties:
+        * city
+        * country
+        * elevation
+        * ep_style_location_string
+        * isLocation
```
chriswmackey: Remove `isLocation`

ladybug/stat.py
```diff
@@ -23,36 +23,42 @@
...
+        * monthly_wb_temp_range_050
+        * monthly_found
+        * standard_pressure_at_elev
+        * monthly_wind_conditions
+        * monthly_ws_avg
+        * monthly_wind_dirs
+        * monthly_clear_sky_conditions
+        * monthly_tau_beam
+        * monthly_tau_diffuse
+        * isStat
```
chriswmackey: Remove `isStat`

ladybug/wea.py
```diff
@@ -49,6 +49,18 @@ class Wea(object):
...
+
+    Properties:
+        * datetimes
+        * direct_normal_irradiance
+        * diffuse_horizontal_irradiance
+        * direct_horizontal_irradiance
+        * global_horizontal_irradiance
+        * header
+        * hoys
+        * isWea
```
chriswmackey: Remove `isWea`

ladybug/dt.py
santiagogaray: Done!

santiagogaray:
@chriswmackey 

Thanks for the detailed review Chris. I think I addressed all the edits but let me know if I missed something or if you have additional comments!

chriswmackey:
This is wonderful @santiagogaray . Thank you, again for doing this. I'll talk to Mostapha later this week to discuss the next best steps for updating documentation and, soon after I pass a big deadline on Nov 8, I will add some new modules to this repo here that show fully how we plan to make cross-platform climate graphics (opening the door for the other project we talked about).

Merged!

santiagogaray:
Sounds good @chriswmackey. I'm very happy I've been able to help in the documentation of these packages and I really look forward to keep helping on these while it makes sense to you guys.  Also happy to get into the big project when things are ready for it!

PR #184, opened 2019-11-12
feat(futil): Add copy_file_tree and remove bat_to_sh

chriswmackey:
As per our recent discussion, @mostapharoudsari , I want to remove this bat_to_sh function before any other parts of Ladybug Tools make use of it (right now, there's nothing using it and it's only here because it was originally a part of the honeybee.fuitl module that this was derived from).

I am also adding a copy_file_tree method, which was useful within the installer component that I just wrote and I think will have some usefulness for us down the line.

Lastly, there are some docstring improvements.

### DIFF
--- a/ladybug/futil.py
+++ b/ladybug/futil.py
@@ -6,6 +6,7 @@ import os
 import shutil
 import zipfile
 import sys
+from distutils import dir_util
 
 if (sys.version_info < (3, 0)):
     import urllib2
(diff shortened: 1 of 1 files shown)


ladybug/futil.py
```diff
@@ -146,32 +167,45 @@ def copy_files_to_folder(files, target_folder, overwrite=True):
     return [os.path.join(target_folder, os.path.split(f)[-1]) for f in files]
 
 
-def bat_to_sh(file_path):
-    """Convert honeybee .bat file to .sh file.
+def copy_file_tree(source_folder, dest_folder, overwrite=True):
```
mostaphaRoudsari: Couldn't we just use `distutils.dir_util.copy_tree` for this?

https://docs.python.org/2/distutils/apiref.html#distutils.dir_util.copy_tree

chriswmackey: You are right. For some reason, I thought that `dir_util.copy_tree` had issues in IronPython but I just tested it and it worked. Maybe it was just an issue with Rhino 5's IronPython. Ok. I'll just remove it from this PR>

PR #185, opened 2019-12-10
fix(sunpath): fix daylight saving period

mostaphaRoudsari:
1. fix daylight saving period
2. add a new method to use calculate sun from minute of the year
3. update Sun class to use ladybug.geometry

### DIFF
--- a/ladybug/sunpath.py
+++ b/ladybug/sunpath.py
@@ -1,8 +1,10 @@
 # coding=utf-8
 from __future__ import division
+import ladybug
 
 from .location import Location
 from .dt import DateTime
+from .analysisperiod import AnalysisPeriod
 from ladybug_geometry.geometry3d.pointvector import Vector3D
 
 import math
@@ -11,7 +13,6 @@ import sys
 if (sys.version_info > (3, 0)):  # python 3
     xrange = range
 
-import ladybug
 try:
     import sunpathplus as plus
 except ImportError as e:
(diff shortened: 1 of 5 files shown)


ladybug/sunpath.py
```diff
@@ -114,11 +115,23 @@ def is_leap_year(self, value):
         """set sunpath to be calculated for a leap year."""
         self._is_leap_year = bool(value)
 
+    @property
+    def daylight_saving_period(self):
+        return self._daylight_saving_period
```
chriswmackey: You are missing a docstring here.

ladybug/sunpath.py
```diff
@@ -114,11 +115,23 @@ def is_leap_year(self, value):
...
+
+    @daylight_saving_period.setter
+    def daylight_saving_period(self, value):
+        if value:
+            assert isinstance(value, AnalysisPeriod), \
+                'Daylight saving period should be an AnalysisPeriod not %s' % type(value)
+            
+        self._daylight_saving_period = value
```
chriswmackey: Am I correct in understanding that the hour value of the AnalysisPeriod is used here? So the change to daylight savings time isn't fixed at 2 AM? If so, I like the flexibility of that feature but you should definitely make a note of it in the docstring since I think that most people probably won't know to set the hour to 2AM if they want an accurate depiction of daylight savings time.

mostaphaRoudsari: Good point. I left it as is for now since AnalysisPeriod is not really fit for this purpose. If we set the starting hour to 2 am and the end hour to another hour it will mean to apply it only for the hours between 2 am to end hour which is not what we really want. That's why I left it for all the hours. That being said the hours that will be wrong are all during the night and so the sun position should not really matter much in that case.

PR #187, opened 2019-12-23
feat(config, logger): Add config and logging modules

chriswmackey:
I added this mostly so that we use a consistent location to download EPW files to. I have verified that this module will work well on both Windows and Unix-based systems.

### DIFF
--- a/ladybug/__init__.py
+++ b/ladybug/__init__.py
@@ -3,22 +3,28 @@ import ladybug._datacollectionbase
 import ladybug.datacollection
 import ladybug.datacollectionimmutable
 
+from ladybug.logutil import get_logger
+
 import sys
 import importlib
 import pkgutil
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,7 +0,0 @@
-FROM python:2.7
```
mostaphaRoudsari: Why are we deleting the docker file?

chriswmackey: Ah sorry. I just saw that none of the other repos had it and I wasn't sure why it was there. I will add it back. Do we need to add a Dockerfile to the other repos?

mostaphaRoudsari: Eventually yes. We want to have a docker image for each of the libraries that one can pull down and use. We are adding them one by one.

chriswmackey: Understood. Thank you for clarifying.

PR #191, opened 2020-01-05
fix(designday): Refactor the DesignDay class with best practices

chriswmackey:
This commit is trying to nip in the bud what I know will be a major naming issue in the future. Right now, the DesignDay object has properties for `ep_style_string`, which manages to be both verbose and yet also vague. Particularly if/when the epjson format replaces the idf format, it will be tough to know whether the "ep_style" refers to an idf or epjson. So I'm replacing all instances of this with `to_idf()` and `from_idf()` as we have it in honeybee_energy.

I'm also implementing some better practices taken from `honeybee_energy`, including the adding of `duplicate` methods along with `==`, `!=`, and `hash` methods.

Lastly, I am trying to use better word choice for the dict keys of the design day. knowing that this is my last chance to change them without breaking `honeybee_energy` or pydantic schema documentation.


chriswmackey:
FYI, I am also making the `location` key within the `DesignDay` schema optional now since I know it adds unnecessary extra redundancy in the `honeybee_energy` workflows.  The only reason why it's there is to help interpret the SkyCondition of the design day if people want to know what the solar irradiance is like for that day.

chriswmackey:
So this turned into a full-on refactor of the DesignDay class and its sub-classes. But everything that I have implemented is just bringing the `DesignDay` up to the standards that we are following in `honeybee_energy` so your review doesn't have to be too thorough, @mostaphaRoudsari . I mostly just want you to be aware that I'm applying these best practices of `honeybee_energy` to the `DesignDay` and the `Location` classes.

ladybug/designday.py
```diff
@@ -1580,21 +1682,33 @@ def radiation_values(self, location, timestep=1):
...
-            analysis_period)
+    def __key(self):
+        """A tuple based on the object properties, useful for hashing."""
+        return (self._month, self._day_of_month, self._tau_b, self._tau_d,
+                self._daylight_savings)
+
+    def __hash__(self):
+        return hash(self.__key())
+
+    def __eq__(self, other):
```
mostaphaRoudsari: Two comments:

1. This module is so long. Can we break it down into smaller files? It will be hard to maintain.
2. Do you have a reference for using hash for comparison methods?

mostaphaRoudsari: Also it might be a good idea to cache the hash when possible. Specially for immutable objects. Here is how datetime module does it:

https://github.com/python/cpython/blob/96b1c59c71534db3f0f3799cd84e2006923a5098/Lib/datetime.py#L770-L772

ladybug/location.py
```diff
@@ -177,28 +192,17 @@ def meridian(self):
 
     def duplicate(self):
         """Duplicate location."""
-        return Location(self.city, self.state, self.country,
-                        self.latitude, self.longitude, self.time_zone, self.elevation,
-                        self.station_id, self.source)
+        return self.__copy__()
 
-    @property
-    def ep_style_location_string(self):
```
mostaphaRoudsari: This is a breaking change. My suggestion is to phase it out in the next couple of releases and meanwhile add a DeprecationWarning.

mostaphaRoudsari: Here is an example in the core library:

https://github.com/python/cpython/blob/822922af90f389b20166f0cd6f5b7c76fd6d42d9/Lib/smtpd.py#L188-L198

ladybug/designday.py
chriswmackey: I agree that is a good idea but `DesignDays` and `Locations` are mutable and I think they really need to be so this suggestion isn't applicable here. Furthermore, most of the objects in `honeybee_energy` are mutable but they are also lockable. It's possible that we can cache the hash while it's locked but I wonder if all the acrobatics that are necessary to reset this cache when it's unlocked will negate the performance savings of caching. Really, the only place in our libraries where this is definitely applicable is `ladybug_geometry` since all geometry objects are immutable. But those don't possess any `__hash__` or `__eq__` functionality right now. It's probably a good idea to add this, though so I opened an issue and I made a note about caching the hash there: 
https://github.com/ladybug-tools/ladybug-geometry/issues/80

ladybug/location.py
chriswmackey: What other code is using this right now that would be broken? It's not used in ladybug_grasshopper right now and I am sending a PR to honeybee_energy that will ensure things keep working over there after the change. Is it being used for the Dynamo plugin?  I would rather make this change now before it's actually used in a stable release of a plugin and people are really relying on it.

BTW, if you are just worried about the few users we have of the pure Python libraries right now, I think these people would rather know sooner than later that this isn't supported.

chriswmackey: I'm also happy to change this PR to be a `refactor` rather than a `fix`, which should trigger a new breaking change release.

chriswmackey:
Thanks for the comments, @mostaphaRoudsari .  To the point about breaking the module up, I'd say that this is still smaller than the `epw` module and, like the `epw` module, it's meant to be for a single file type. So breaking up this module would be a bit like breaking the `EPWFields` and `EPWField` classes out of the `epw` module. BTW, I'd be ok breaking up the `epw` module this way (I think it would be fine if those `Filed` classes existed in a separate hidden module since they don't really have a use on their own). But I'm just saying we should be consistent in our treatment of the modules.

To this point, I would actually be ok breaking the DDY class out of the `designday` module. It wouldn't change the overall length of the module by much (it would shave off ~200 lines). But `DesignDay` objects have a usefulness outside of the `DDY` object. The same cannot be said for the `DryBulbCondition` class (or any of the other "condition" classes), which exists only as a means to support the `DesignDay` class.

chriswmackey:
Looking it over, I think it makes sense to break out the `DDY` class into another `ddy` module. I'll do this now. Let me know if you have any thoughts on breaking up the `epw` module.

ladybug/designday.py
mostaphaRoudsari: My main question about the idea of using hash for comparison remains. How did you find this to be a good practice? Did you find this implementation in one of the standard libraries or did you read about it somewhere?

ladybug/location.py
mostaphaRoudsari: > It's not used in ladybug_grasshopper right now and I am sending a PR to honeybee_energy that will ensure things keep working over there after the change. Is it being used for the Dynamo plugin?

It is about the library itself not about who else is using it. We should do it right by using DeprecationWarning and deprecate it in a couple of releases.

>  I would rather make this change now before it's actually used in a stable release of a plugin and people are really relying on it.

I didn't say we shouldn't make the change. I just think we should do it right. You can add the new one and keep the backwards compatibility for a number of releases with DeprecationWarning. That will give other developers a chance to update their code to the latest version.

ladybug/designday.py
chriswmackey: I don't use `__hash__` for comparison. `__hash__` only gets used by the native Python `set()` and `hash()` operators (not `==` or `!=`).

It's true that I use a `__key` for all 4 of these operators, though, so that there's only one place where all of the unique properties of the object are defined. This way, I don't have to worry about a case where `set()` evaluates two different objects but `==` tells me that the two objects are equivalent.  This particular suggestion comes from the most up-voted suggestion on StackOverflow:

https://stackoverflow.com/questions/2909106/whats-a-correct-and-good-way-to-implement-hash

ladybug/location.py
chriswmackey: Ok, well, if the point is to practice "doing it right", we need some type of system to remind us to actually implement the breaking change because I know that I'm most likely going to forget about it otherwise. Do we want to keep the reminders as github issues (I'll likely forget to check them here unless maybe we are able to set a reminder about them through Zenhub)? Or Basecamp?

There are also some questions to figure out like "how long does the deprecation warning need to be around before we're able to implement the change?"

I was hoping we wouldn't have to worry about setting up this whole process until its really needed but, if you want to figure this out now, we can do that.

mostaphaRoudsari: > If it's only this particular breaking change that you are worried about in this PR, I can just take this deletion out until we can discuss these things.

Let's talk before you remove them. I feel you are overthinking this. It is a common practice and I didn't come up with it. You can just look up the best practices for code deprecation.

ladybug/designday.py
mostaphaRoudsari: Thanks but the example is different from your implementation. In the example the `__key` is a tuple of attributes. In your case you have nested hashing for each item in the tuple. Is that necessary?

chriswmackey: I see your question now. Yes, it's necessary to hash it because nested tuples cannot be hashed. By hashing them within the `__key`, we are able to ensure everything is hash-able up the whole nested hierarchy.

ladybug/location.py
chriswmackey: I know it's a common practice and it's something that we'll definitely need to do regularly after the official release (much like the use of development branches that we aren't really using right now). But, like the use of the development branch, there are a couple of rules and best practices about them that, if not followed, make it easy to ignore.

chriswmackey: I removed this change from this PR and I set up an issue tag for the changes that we should probably give a deprecation warning about before the next plugin release: https://github.com/ladybug-tools/ladybug/labels/for%20the%20next%20release

I'll aim to send a "deprecation warning PR" at some point soon.

ladybug/designday.py
chriswmackey: I researched this a bit more and it turns out that [hash collisions](https://preshing.com/20110504/hash-collision-probabilities/) are, in fact, a thing that could happen. But I think it follows from logic that the probability of encountering these with the nested hashes here isn't any larger than the probability of encountering such a collision with a non-nested hash. And I have some faith that the hashing function under the hood of modern Python installations has a low enough probability of collision that we won't encounter it (hopefully, those are not famous last words!).

PR #199, opened 2020-01-07
style(pep8): Remove all `is True` and `is False` cases

chriswmackey:
The PEP8 police should appreciate this commit. It's purely a style commit that removes all situations of `is True` and `is False` from the library, replacing them with the correct PEP8 version.


chriswmackey:
There's really nothing to review here @mostaphaRoudsari . I just requested it because I wanted you to be aware of the change.

Issue #192, opened 2020-01-06
Replace all use of hasattr with isinstance across the library and get rid of all "is" properties


chriswmackey:
We know that the original reason why `hasattr` is used throughout this library is because it was a workaround to get the library to run in Dynamo and that the recommended practice is to use `isinstance` instead. Given that we agreed the core library shouldn't be burdened with non-recommended practices in order to accommodate workarounds for specific interfaces, we should remove all of the properties starting with `is` before the next stable release of the plugins and replace the use of `hasattr` with `isinstance`.

I can confirm that the current `ladybug_grasshopper` components do not use `hasattr` right now and so implementing this change should not break them.

mostaphaRoudsari:
I did a quick search and it looks like all the current usage of `hasattr` is meaningful. We can probably close this issue.

chriswmackey:
The main reason why I opened this issue is because of all the "is" properties like this:
https://github.com/ladybug-tools/ladybug/blob/master/ladybug/location.py#L141-L144
and this:
https://github.com/ladybug-tools/ladybug/blob/master/ladybug/datatype/temperature.py#L47-L50

Once we get rid of those methods, I think we can close this issue. Am I clear to delete all of them or do we need to worry about breaking Dynamo and/or give a deprecation warning?

mostaphaRoudsari:
Ah! I see. I searched for `is_.*` and didn't find any. The Dynamo packages are not connected to source here so you can go ahead and remove them without deprecation. Let's hope no one else is using them already.

chriswmackey:
That's great to hear.  I should have asked earlier as I didn't realize Dynamo wasn't dependent on this repo at all. I'll send a PR now.

Few things are as satisfying as deleting bad code 🙂 !

chriswmackey:
Also, the reason why the search didn't turn up anything is that these old methods are all using cammelCase. Even all the more reason to delete it!

chriswmackey:
Addressed via https://github.com/ladybug-tools/ladybug/pull/205

PR #205, opened 2020-01-24
fix(is): Remove all isClass properties

chriswmackey:
As established [here](https://github.com/ladybug-tools/ladybug/issues/192), this is not good practice and so it is being removed.


chriswmackey:
There really isn't much to review here but I'm doing it as a formality and to see if it affects the CLA app.

Issue #196, opened 2020-01-06
Add `__hash__` and `__eq__` functionality to all classes


chriswmackey:
It would be nice to be able to use `==`, `!=` and `hash` operators with the objects in this library. In order to do this, we need to add some hidden methods to each of the geometry classes. A sample of these methods can be found here:
https://github.com/ladybug-tools/honeybee-energy/blob/master/honeybee_energy/material/opaque.py#L513-L525

It's just that, for the `__key`, we should use whatever properties make the object unique.

javier123454321:
Hi, I started with this process of adding the a __key(), __hash__(), __eq__(), and __ne__() functions to all classes. 

So far I have done it for Location, STAT, Sunpath, Sun, and Wea classes, and adding unit tests to verify this functionality. One thing to consider is that for some classes, if an object is part of the keys, you need to evaluate the keys as strings. 

This would be my first contribution to this project. I have not finished it, but I am wondering if I should submit a PR for the work I have so far. At any rate, I would love to know how to proceed here, and happy to help.

chriswmackey:
Thank you for starting to tackle this @javier123454321 . As a general guideline, I would recommend starting with simpler classes like Location, AnalysisPeriod, Color, etc. The `==` operator will be most helpful on these simpler classes and, as you can can already tell, implementing the operators on complex objects like Wea or STAT presents some issues.

To clarify further, evaluating the sub-objects as strings is NOT the correct way to implement the operators when the properties of an object are other objects. The string representation of an object usually doesn't include everything that is needed define that object. Instead, the correct way to implement it is by hashing the sub-objects as you see here on the DDY class:
https://github.com/ladybug-tools/ladybug/blob/master/ladybug/ddy.py#L200

This requires you to have actually implemented a `__hash__` operator on these sub-objects, hence, why I said that it's better to start with simple objects and work your way up to the complex ones.  I'd recommend sending a PR that is just for a few simple objects first. Then, we can verify that those operators are correct, merge that in, and you can use them to write the operators for more complex classes.

javier123454321:
It makes sense to tackle it through hashes rather than the string version of the keys. I sent in a small pr just tackling analysisperiod. If this is a go, I'll do more classes in this way.

chriswmackey:
I think we can call this one pretty much done.  We have everything needed to check the equality of data collections, which is the main use case.

PR #206, opened 2020-02-09
#196 add `  hash  ` and `  eq  ` functionality to all classes

javier123454321:
Updated the __eq__() method to check for the keys. I only used the parameters that were currently being evaluated in the existing __eq__() method.

### DIFF
--- a/ladybug/analysisperiod.py
+++ b/ladybug/analysisperiod.py
@@ -426,17 +426,6 @@ class AnalysisPeriod(object):
         end_doy = sum(self._num_of_days_each_month[:end_time.month-1]) + end_time.day + 1
         return list(range(start_doy, end_doy))
 
-    def __eq__(self, other):
-        """Whether the inputs match between two analysis periods."""
-        if isinstance(other, self.__class__):
-            if (self.st_time, self.end_time, self.timestep, self.is_leap_year) == \
-                    (other.st_time, other.end_time, other.timestep, other.is_leap_year):
-                return True
-        return False
-
-    def __ne__(self, other):
-        return not self.__eq__(other)
-
     def __len__(self):
         """Number of hours of the year.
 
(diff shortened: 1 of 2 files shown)


ladybug/analysisperiod.py
```diff
@@ -463,3 +452,16 @@ def __repr__(self):
              self.end_time.month, self.end_time.day,
              self.st_time.hour, self.end_time.hour,
              self.timestep)
+
+    def __key(self):
+        return(self.st_time, self.end_time, self.timestep, self.is_leap_year)
```
chriswmackey: I think you need to `hash()` the datetime objects in order for the `__hash__` method to work correctly. We'd know for sure if you add a test like this:
```
a_period = AnalysisPeriod()
hash(a_period )
```
And you get an exception.

tests/analysisperiod_test.py
```diff
@@ -180,3 +180,11 @@ def test_0_end_hour():
...
     assert ap.end_hour == 0
     assert len(ap) == 24
+
+def test_equality():
+    ap_one = AnalysisPeriod()
+    ap_two= AnalysisPeriod()
+    ap_three = AnalysisPeriod(st_month=2, end_month=6)
+
+    assert ap_one == ap_two
+    assert ap_one != ap_three
```
chriswmackey: This test should be a bit more robust. Here is a good template for you to follow when writing tests for these operators:
https://github.com/ladybug-tools/ladybug-geometry/blob/master/tests/pointvector3d_test.py#L34-L45

chriswmackey:
Also, you will need to digitally sign the contributors license agreement (CLA) before we can merge.

javier123454321:
Ok so It seems that the object is hashing fine when analyzing the keys of the analysisperiod object, and the new test cover that case. It works fine as is.

As far as the contributor agreement, I did the changes without an email attached to my git, therefore making the 3 original commits to come from an unknown id to github. I tried to ammend the commits with an email attached, but the new commits with the email that signed the CLA show as a different history. (this is getting to messy git territory that I am unfamiliar with) I tried to clean it as best I could.

chriswmackey:
Thanks for adding the tests and I can see why it all passes now. I forgot that the ladybug datetime inherits from the native python datetime, which already has a `__hash__` function on it. 

All looks good, then. We can merge this in as soon as you sign the CLA.

javier123454321:
Finally rebased the initial commits to change the author and put the right email in the history. This overwrote the original history, but is now all done, and the CLA is finally signed with the updated changes. I will continue with this task for another class using this framework as reference.

PR #210, opened 2020-03-07
feat(hourlyplot): Add module for HourlyPlot

chriswmackey:
This commit adds a module to create geometry of hourly plots for any hourly DataCollection.

It also fixes a few bugs and adds a few small features to other modules that pertain to the hourly plot.

@mostaphaRoudsari ,
I think it would be helpful if you could review this PR since it will serve as a template for many of the other modules that produce climate graphics.
I still have to add tests but I've been testing it in Rhino/Grasshopper with a wide variety of cases. It's working well so I don't anticipate many changes as I write the unit tests.

### DIFF
--- a/ladybug/hourlyplot.py
+++ b/ladybug/hourlyplot.py
@@ -1,0 +1,510 @@
+# coding=utf-8
+"""Module for visualization of houlry data collections."""
+from __future__ import division
+
+from .analysisperiod import AnalysisPeriod
+from .datacollection import HourlyContinuousCollection, HourlyDiscontinuousCollection
+from .graphic import GraphicContainer
+from .legend import LegendParameters
+
+from ladybug_geometry.geometry2d.pointvector import Point2D, Vector2D
+from ladybug_geometry.geometry2d.line import LineSegment2D
+from ladybug_geometry.geometry2d.polygon import Polygon2D
+from ladybug_geometry.geometry2d.polyline import Polyline2D
+from ladybug_geometry.geometry2d.mesh import Mesh2D
+from ladybug_geometry.geometry3d.pointvector import Point3D, Vector3D
+from ladybug_geometry.geometry3d.line import LineSegment3D
+from ladybug_geometry.geometry3d.plane import Plane
+from ladybug_geometry.geometry3d.polyline import Polyline3D
+from ladybug_geometry.geometry3d.mesh import Mesh3D
+
+
+class HourlyPlot(object):
+    """Object for visualization of houlry data collections.
+
+    Args:
+        data_collection: A HourlyContinuousCollection or HourlyDiscontinuousCollection
+            which will be used to generate the hourly plot.
+        legend_parameters: An optional LegendParameter object to change the display
+            of the HourlyPlot (Default: None).
+        base_point: An optional Point3D to be used as a starting point to generate
+            the geometry of the plot (Default: (0, 0, 0)).
+        x_dim: An optional number to set the X dimension of the mesh cells (Default: 1).
+        y_dim: An optional number to set the Y dimension of the mesh cells (Default: 4).
+        z_dim: An optional number to set the Z dimension of the entire chart. This
+            will be used to make the colored_mesh3d of the chart vary in the Z
+            dimension according to the data. The value input here should usually be
+            several times larger than the x_dim or y_dim in order to be noticable
+            (e.g. 100). If 0, the colored_mesh3d will simply be flat. (Default: 0).
+
+    Properties:
+        * data_collection
+        * legend_parameters
+        * base_point
+        * x_dim
+        * y_dim
+        * z_dim
+        * colored_mesh2d
+        * colored_mesh3d
+        * legend
+        * chart_border2d
+        * chart_border3d
+        * hour_lines2d
+        * hour_lines3d
+        * hour_label_points2d
+        * hour_label_points3d
+        * hour_labels
+        * month_lines2d
+        * month_lines3d
+        * month_label_points2d
+        * month_label_points3d
+        * month_labels
+        * title_text
+        * lower_title_location
+        * upper_title_location
+        * analysis_period
+        * values
+        * colors
+    """
+    __slots__ = ('_data_collection', '_base_point', '_x_dim', '_y_dim', '_z_dim',
+                 '_num_y', '_num_x', '_container', '_colored_mesh2d', '_colored_mesh3d',
+                 '_hour_points', '_hour_text', '_month_points', '_month_label_points',
+                 '_month_text')
+
+    HOUR_LABELS = (0, 6, 12, 18, 24)
+
+    def __init__(self, data_collection, legend_parameters=None, base_point=Point3D(),
+                 x_dim=1, y_dim=4, z_dim=0):
+        """Initialize hourly plot."""
+        # check the input objects
+        acceptable_colls = (HourlyContinuousCollection, HourlyDiscontinuousCollection)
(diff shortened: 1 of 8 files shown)


chriswmackey:
BTW, this is the corresponding PR for ladybug-grasshopper:
https://github.com/ladybug-tools/ladybug-grasshopper/pull/21

mostaphaRoudsari:
Hey @chriswmackey! Thank you for the PR. I will review this tomorrow morning.

chriswmackey:
Awesome. Thank you for letting me know your plan. The tests are all done now so, once we factor in your feedback, we should be good to merge.

Also, so that you have something visual to go off of, here are a few screenshots:
image

image

image

image

image


image

ladybug/hourlyplot.py
```diff
@@ -0,0 +1,424 @@
...
+from ladybug_geometry.geometry2d.pointvector import Point2D, Vector2D
+from ladybug_geometry.geometry2d.line import LineSegment2D
+from ladybug_geometry.geometry2d.polygon import Polygon2D
+from ladybug_geometry.geometry2d.polyline import Polyline2D
+from ladybug_geometry.geometry2d.mesh import Mesh2D
+from ladybug_geometry.geometry3d.pointvector import Point3D, Vector3D
+from ladybug_geometry.geometry3d.line import LineSegment3D
+from ladybug_geometry.geometry3d.plane import Plane
+from ladybug_geometry.geometry3d.polyline import Polyline3D
+from ladybug_geometry.geometry3d.mesh import Mesh3D
```
mostaphaRoudsari: Not related to this PR but just seeing this it might be a good idea to add the commonly used objects to `__init__` file to make it easier to access them for a user.

```python
from ladybug_geometry.geometry2d import Point2D, Vector2D
```

ladybug/hourlyplot.py
```diff
@@ -0,0 +1,424 @@
...
+            of the HourlyPlot (Default: None).
+        base_point: An optional Point3D to be used as a starting point to generate
+            the geometry of the plot (Default: (0, 0, 0)).
+        x_dim: An optional number to set the X dimension of the mesh cells (Default: 1).
+        y_dim: An optional number to set the Y dimension of the mesh cells (Default: 4).
+        z_dim: An optional number to set the Z dimension of the entire chart. This
+            will be used to make the colored_mesh3d of the chart vary in the Z
+            dimension according to the data. The value input here should usually be
+            several times larger than the x_dim or y_dim in order to be noticable
+            (e.g. 100). If 0, the colored_mesh3d will simply be flat. (Default: 0).
```
mostaphaRoudsari: We had to make a similar decision in buttefly before and it looks like it is easier for people to give you the final desired size than calculating the cell size. At the same time the cell size approach is much better for comparative studies where you want to keep the size the same. We might want to consider a `classmethod` to give them both options.

ladybug/hourlyplot.py
```diff
@@ -0,0 +1,424 @@
...
+        * legend
+        * chart_border2d
+        * chart_border3d
+        * hour_lines2d
+        * hour_lines3d
+        * hour_label_points2d
+        * hour_label_points3d
+        * hour_labels
+        * month_lines2d
+        * month_lines3d
```
mostaphaRoudsari: It is a design call but it might be cleaner to consider some of these properties as methods and pass `3D=True/False` as an argument.

I like the overall approach here very much.

ladybug/hourlyplot.py
```diff
@@ -0,0 +1,424 @@
...
+        * month_label_points2d
+        * month_label_points3d
+        * month_labels
+        * title_text
+        * lower_title_location
+        * upper_title_location
+        * analysis_period
+        * values
+        * value_colors
```
mostaphaRoudsari: What about `colors`?

ladybug/hourlyplot.py
```diff
@@ -0,0 +1,424 @@
...
+    """
+    __slots__ = ('_data_collection', '_base_point', '_x_dim', '_y_dim', '_z_dim',
+                 '_num_y', '_num_x', '_container', '_colored_mesh2d', '_colored_mesh3d',
+                 '_hour_points', '_hour_text', '_month_points', '_month_label_points',
+                 '_month_text')
+
+    HOUR_LABELS = (0, 6, 12, 18, 24)
```
mostaphaRoudsari: I think this will be something that people want to have access to. Specially now that you are providing them a very well-designed plot with a lot of flexibility.

ladybug/hourlyplot.py
```diff
@@ -0,0 +1,424 @@
...
+
+    @property
+    def y_dim(self):
+        """A number for the Y dimension of each cell of the hourly plot."""
+        return self._y_dim
+
+    @property
+    def z_dim(self):
+        """A number for the Z dimension of each cell of the hourly plot."""
+        return self._z_dim
```
mostaphaRoudsari: Not that it should be like that but I was expecting to be able to set these values without creating a new chart. Basically if the data stays the same plot should give me the ability to play around with these input values to get the right visualization without recreating the object.

Curious to know the reasoning behind making them read-only?

mostaphaRoudsari:
Just saw the graphics. I 😍 the flexibility in creating beautiful legends. We should include some of these graphics in our release notes.

Also adding some of these graphics to README.md can make them more human-friendly.

ladybug/hourlyplot.py
chriswmackey: Ok. I opened an issue
https://github.com/ladybug-tools/ladybug-geometry/issues/100

ladybug/hourlyplot.py
chriswmackey: Particularly because the highest value in the data set is going to be different for different data types (eg. Temperature vs. Illuminance), I think the total z dimension is going to be a much better default. But I'll add a classmethod along the lines of what you suggest with a z_dim_per_unit.

ladybug/hourlyplot.py
chriswmackey: I was considering something like that but I think the properties just resulted in cleaner looking code on the Grasshopper component end. It might be worth revisiting this later.

ladybug/hourlyplot.py
chriswmackey: Done.

ladybug/hourlyplot.py
chriswmackey: The only reason why I didn't do that originally is that I didn't want people confusing it with the `legend_parameter.colors`.  But maybe this really isn't as confusing as I thought. Ok, I'll change it.

ladybug/hourlyplot.py
chriswmackey: The main reason for it is because everything you see running in `__init__` would have to be re-run every time that one of these properties changes. This includes generating a new graphic container object based on new min and max points that have now changed because the dimensions have changed.  It also prevents me from caching certain hidden variables that get reused in multiple properties.

Bottom line is that it's just a lot of extra code to enable flexibility that I'm not sure we're actually going to use. If we find situations that we really want this feature, then we can sit down and add in all of the extra code to enable it. But it just didn't seem worth it right now.

ladybug/hourlyplot.py
chriswmackey: That's why I made it public. If people really want to change it everywhere, they can overwrite `HOUR_LABELS `. But I'll throw in a method to this class that can generate custom hour lines and text labels based on any list of labels that the user inputs. This should satisfy your comment here.

chriswmackey: Done!

PR #214, opened 2020-03-16
feat(sunpath): Refactor sunpath to use ladybug_geometry instead of plus

chriswmackey:
This commit switches the Ladybug sunpath to use ladybug_geometry to generate all of its graphics. This effectively allows the production of highly customized geometry using only Python.

This commit also adds a module for generating compass graphics, which are intended to be used by the sunpath as well as graphics like the wind rose that @saeranv is working on.

Lastly, I found and fixed quite a lot of bugs and made several enhancements to the sun path including:

* Time zones for a lot of EPW locations were being "autocorrected" to solar time. Notably, at least half of Chinese sunpaths did not have the right time zone from the EWP file. Instead of this, I made sure that a user-specified (or epw-specified) time zone always takes precedence but, if a user does not specify a time zone, it will default to the solar time zone based on the longitude.
* Fixed a math domain error that was occurring whenever someone requested solar vectors at precise solar noon.
* Fixed a math domain error that was occurring whenever the north or south pole was used as a location.
* Fixed a syntax bug in the calculation of atmospheric refraction. Essentially, multiple lines of code were not being included in the refraction factor calculation. I am guessing this is leftover because this Sunpath class was one of the first classes Mostapha tried to write with a 79 characters-per-line limit and the correct syntax for multi-line statements wasn't yet ingrained.
* Fixed the is_daylight_saving_hour method to correctly interpret the st_hour and end_hour of the analysis period as the time of the day when daylight saving starts/ends. Before, if a user had messed with these start and end hours, they would have gotten a sunpath jumping in and out of daylight saving every day.
* Fixed the application of daylight savings to move the clocks forward (it was incorrectly moving the clock backwards previously)
* Changed ladybug to use degrees for everything (no radians exposed on the public methods or properties). Most notably, this means that the __init__ of Sun takes altitude and azimuth in degrees. So we can now say that all of the ladybug tools libraries use degrees for angle inputs except for ladybug_geometry.
* Simplified a lot of the conversions that we happening back and forth between degrees and radians. Now, for most internal calculations we use radians and for all external properties/methods we use degrees.
* Flattened the structure of some methods nested in other methods since some of these methods were long and it was making it hard to read when one method ended and the next one began.
* Added a bunch of docstrings.
* Added __repr__ for the sunpath.

@mostaphaRoudsari ,
This PR is ready for review, though I still have to add plenty of tests. I've been testing it heavily on the Rhino end, though, and I will send the updated Ladybug[+] component in a PR shortly.


chriswmackey:
Here is the corresponding PR in ladybug_grasshopper:
https://github.com/ladybug-tools/ladybug-grasshopper/pull/22

ladybug/sunpath.py
```diff
@@ -80,34 +88,63 @@ def from_location(cls, location, north_angle=0, daylight_saving_period=None):
...
+        return self._time_zone
 
-        # update time_zone
-        if abs((value / 15.0) - self.time_zone) > 1:
-            # if time_zone doesn't match the longitude update the time_zone
-            self.time_zone = value / 15.0
+    @time_zone.setter
+    def time_zone(self, tz):
+        self._time_zone = round(self.longitude / 15) if tz is None else float(tz)
+        assert -12 <= self._time_zone <= 12, \
```
mostaphaRoudsari: If we agree to change maximum to 14 then we have to update it here too. We should also update it in all the docstrings. 

```suggestion
        assert -12 <= self._time_zone <= 14, \
```

ladybug/sunpath.py
```diff
@@ -117,38 +154,47 @@ def is_leap_year(self, value):
...
-            day: An integer between 1-31
-            hour: A positive number between 0..23
+            month: An integer between 1 and 12.
+            day: An integer between 1 and 31.
+            hour: A positive number between 0 and 23.
             is_solar_time: A boolean to indicate if the input hour is solar time.
-                (Default: False)
+                Note that this input is redundant and should not be used if the
+                time_zone of the sun path has already been set to a decimal
+                solar time. (Default: False)
```
mostaphaRoudsari: Can we think of a better way to implement this?

ladybug/sunpath.py
```diff
@@ -344,84 +412,328 @@ def calculate_sunrise_sunset_from_datetime(self, datetime, depression=0.833,
                                    leap_year=self.is_leap_year)
             }
 
+    def analemma_suns(self, time, daytime_only=False, is_solar_time=False):
```
mostaphaRoudsari: I reviewed up to here. Will review the rest after the call.

ladybug/compass.py
```diff
@@ -0,0 +1,259 @@
...
+        """Get a Arc2D for the inner circle of the compass.
+        
+        This is essentially a circle with the compass radius.
+        """
+        return Arc2D(self.center, self.radius) 
+
+    @property
+    def all_boundary_circles(self):
+        """Get an array of 3 Arc2Ds for the circles of the compass."""
+        arc2 = Arc2D(self.center, self.radius * (1 + self.spacing_factor * 0.1))
```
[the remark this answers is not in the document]
chriswmackey: I'm ok if you want to open an issue about this on ladybug-geometry but, if you do so, it would be good to know what things you see distinguishing these classes from one another since, if we were to implement it now, it would probably just result in an empty class (either circle or arc) that inherits everything from the other class. It would also be good to know how you would recommend dealing with arcs objects that are circles (eg. should we not allow an arc to be a closed and thus get rid of the [is_circle](https://www.ladybug.tools/ladybug-geometry/docs/ladybug_geometry.geometry2d.arc.html#ladybug_geometry.geometry2d.arc.Arc2D.is_circle) property).

ladybug/dt.py
```diff
@@ -346,10 +346,10 @@ def from_doy(cls, doy, leap_year=False):
         """
         if not leap_year:
             days_until_month = (0, 31, 59, 90, 120, 151, 181,
-                                212, 243, 273, 304, 334, 365)
+                                212, 243, 273, 304, 334, 366)
```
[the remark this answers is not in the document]
chriswmackey: Because it was not allowing me to `__init__` a datetime for 31 Dec thanks to the code that comes after this.

ladybug/location.py
```diff
@@ -140,37 +142,40 @@ def from_idf(cls, idf_string):
...
-        """Location time zone."""
+        """Get or set the location time zone as a number between -12 and + 12."""
         return self._tz
 
     @time_zone.setter
     def time_zone(self, tz):
-        self._tz = 0 if not tz else float(tz)
-        assert -12 <= self._tz <= 12, "Time zone should be between -12.0..12.0"
+        self._tz = round(self._lon / 15) if tz is None else float(tz)
+        assert -12 <= self._tz <= 12, \
```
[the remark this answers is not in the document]
chriswmackey: Oh, wow. You're right!  Because of the shape of the international dateline, there's a significant part of the Pacific ocean that goes past +12:
https://www.timeanddate.com/time/map/
It looks like mainland NZ is still in +12 but some of the islands off the coast that dip into this region. Ok, I will correct this now for both the Location and the Sunpath.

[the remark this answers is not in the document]
chriswmackey: Done!

tests/sunpath_test.py
```diff
@@ -62,3 +66,40 @@ def test_leap_year():
     assert sun.datetime.month == 2
     assert sun.datetime.day == 29
     assert sun.datetime.hour == 11
+
```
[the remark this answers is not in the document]
chriswmackey: While I made a number of bug fixes here, I am not sure that any of them are going to fix the discrepancy with Revit. When I write the tests today, I will check but I'm not too hopeful that the discrepancy is closed since I think the issue is ultimately that Revit's calculations are not as accurate as ours.

ladybug/sunpath.py
chriswmackey: Done!

ladybug/sunpath.py
chriswmackey: I think it's useful to have the two different ways to get solar time:
1. Set this argument to True
2. Set the time zone of the sun path to be `None`.
The first one means that you can still get the solar time even if you init a Sunpath from an EPW location with an integer time zone and the second one is useful for cases where you don't know/don't care about time zones and just need some solar positions that reflect the latitude.

I think I may just remove this part of the docstring, though, since it's not really helpful or correct. It's fine if people set it to True while using a `Sunpath.time_zone` that's set to solar time. It's an extra step that won't change the result but it' not like it breaks or it's incorrect if they set it to True for this case.

chriswmackey: Updated the dcocstring.

ladybug/compass.py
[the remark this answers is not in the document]
mostaphaRoudsari: I assume circle is a subclass of Arc where the angle is always 360 and is_circle always returns true. As simple as that.

ladybug/sunpath.py
```diff
@@ -736,38 +940,37 @@ class Sun(object):
         * data
         * is_during_day
         * sun_vector
+        * sun_vector_reversed
```
mostaphaRoudsari: Why not just leave this to user to reverse the sun vector if needed?

chriswmackey: There are 5-6 places where I need the reversed sun vector to generate different parts of the sunpath geometry or to generate points that get plotted on this geometry. So it definitely seemed worthwhile to store the reversed vector on the `Sun` object rather than regenerating it each time. If you think this may confuse people, I can store it just as a hidden property and use it internally within the sunpath module.

I think I'll merge the PR for now but, if you think it's helpful to keep the property hidden, just let me know here and I'll send another PR that hides it.

ladybug/compass.py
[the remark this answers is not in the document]
chriswmackey: Ok. So it sounds like you just overwrite `__init__` for the circle to not expose the angular cutoffs, I guess that we can also add properties like `area` that don't really have meaning for an arc. Ok, feel free to open an issue about it on ladybug-geometry and I'll get to it when I can.

ladybug/sunpath.py
mostaphaRoudsari: > If you think this may confuse people, I can store it just as a hidden property and use it internally within the sunpath module.

I think it will. Having it removed or hidden will be helpful.

PR #215, opened 2020-03-21
WindRose module

saeranv:
This module represents the windrose. This image gives an idea of the main types of representations:

image

Aspects that are still WIP:
- ~50% of the docstrings
- The last image, show_stack=True, show_zero=True needs to be debugged
- legend parameter needs to reflect actual histogram binning 

@chriswmackey also note that there are three functions:

> `_linspace`
> `bin_histogram_data`
> `bin_histogram_data_circular`

That I've moved out of the WindRose class. These are the three functions I mentioned that might be useful to move into the datacollection. Let me know your thoughts. 

S

### DIFF
--- a/ladybug/windrose.py
+++ b/ladybug/windrose.py
@@ -19,11 +19,139 @@ from ladybug_geometry.geometry3d.polyline import Polyline3D
 from ladybug_geometry.geometry3d.mesh import Mesh3D
 
 
-from math import pi, cos, sin, atan2, log, ceil
+from math import pi, cos, sin, ceil
+
+
+def _linspace(start, stop, num):
+    """Return evenly spaced numbers over a specified interval.
+
+    Returns num evenly spaced samples, calculated over the interval [start, stop].
+    Equivalent to numpy.linspace(0, 5, 6) -> array([0., 1., 2., 3., 4., 5.])
+
+    Args:
+        bin_num: Number of bins
+        bin_range: Tuple representing range of values
+
+    Returns:
+        A list of numbers defining bin intervals across bin range.
+    """
+    try:
+        delta = stop - start
+        return [i * (delta / (num - 1)) + start for i in range(num)]
+    except ZeroDivisionError:
+        return [start]
+
+
+def bin_histogram_data(values, bin_arr, bin_range, key=None):
+    """Compute the histogram from this object's data collection.
+
+    The data is binned inclusive of the lower bound but exclusive of the
+    upper bound for intervals.
+
+    Example of where we lose 3 because of exclusive upper bound:
+    histogram([0, 0, 0.9, 1, 1.5, 1.99, 2, 3], (0, 1, 2, 3))
+    >>> [[0, 0, 0.9], [1, 1.5, 1.99], [2]]
+
+    Args:
+        values: list of numerical data to bin.
+        bin_arr: A list of bin bounds.
+        is_gradient: Optional boolean parameter to pass to nest datas for histogram gradient
+
+    Returns:
+        A list of lists representing the ordered values binned by frequency.
+            Example: histogram([0, 1, 1, 2, 3], [0, 2, 3]) -> [[0, 1, 1], [2]]
+    """
+
+    if key is None:
+        key = lambda v: v
+
+    vals = sorted(values, key=key)
+
+    bin_bound_num = len(bin_arr)
+
+    # Init histogram bins
+    hist = [[] for i in range(bin_bound_num - 1)]
+    bin_index = 0
+    for val in vals:
+        k = key(val)
+        # Ignore values out of range
+        if k < bin_range[0] or k >= bin_range[1]:
+            continue
+
+        # This loop will iterate through the bin upper bounds.
+        # If the value is within the bounds, the lower bound
+        # of the bin_index is updated, and the loop is broken
+        for i in range(bin_index, bin_bound_num - 1):
+            if k < bin_arr[i + 1]:
+                hist[i].append(val)
+                bin_index = i
+                break
+
+    return hist
+
+
+def bin_histogram_data_circular(values, bin_arr, bin_range, key=None):
+    """Bin data into a histogram for cyclical data.
+
+    The data is binned inclusive of the lower bound but exclusive of the
+    upper bound for intervals.
+
+    Example of where we lose 3 because of exclusive upper bound:
+    histogram([0, 0, 0.9, 1, 1.5, 1.99, 2, 3], (0, 1, 2, 3))
+    >>> [[0, 0, 0.9], [1, 1.5, 1.99], [2]]
+
+    Args:
+        values: list of numerical data to bin.
+        bin_arr: A list of bin bounds.
+        is_gradient: Optional boolean parameter to pass to nest datas for histogram gradient
+
+    Returns:
+        A list of lists representing the ordered values binned by frequency.
+            Example: histogram([0, 1, 1, 2, 3], [0, 2, 3]) -> [[0, 1, 1], [2]]
+    """
+
+    if key is None:
+        key = lambda v: v
+
+    vals = sorted(values, key=key)
+
+    bin_bound_num = len(bin_arr) - 1
+
+    # Init histogram bins
+    hist = [[] for i in range(bin_bound_num)]
+    for val in vals:
+        k = key(val)
+
+        # Ignore values out of range
+        if k < bin_range[0] or k >= bin_range[1]:
+            continue
+
+        # This loop will iterate through the bin upper bounds.
+        # If the value is within the bounds, the loop is broken.
+        # Since values at the end of the list can still be binned
+        # into the earlier histogram bars since we have circular
+        # data we don't update the bin_index.
+        for i in range(bin_bound_num):
+            if bin_arr[i] > bin_arr[i + 1]:
+                # If the interval starts data from the end of the list,
+                # split the conditional checks into two to check two
+                # intervals.
+                interval1 = (k < bin_range[1] and k >= bin_arr[i])
+                interval2 = (k < bin_arr[i + 1] and k >= bin_range[0])
+                if interval1 or interval2:
+                    hist[i].append(val)
+                    break
+            else:
+                if k < bin_arr[i + 1]:
+                    hist[i].append(val)
+                    break
+
+    return hist
 
 
 class WindRose(object):
     """Module for visualization of wind data collection by orientation.
+    # TODO: Add rest of args.
     Args:
         data_collection: A HourlyContinuousCollection or HourlyDiscontinuousCollection
             which will be used to generate the windrose.
@@ -32,22 +160,23 @@ class WindRose(object):
         base_point: An optional Point3D to be used as a starting point to generate
             the geometry of the plot (Default: (0, 0, 0)).
         scale: The scale of the wind rose.
+        number_of_directions: The number of orientations to divided the wind rose
+            directions.
 
     Properties:
         * direction_data_collection
         * speed_data_collection
+        * direction_values
+        * speed_values
+        * analysis_period
         * legend
         * legend_parameters
         * base_point
         * radius
         * colored_mesh
-        * chart_border
+        * border
         * orientation_lines
         * frequency_lines
-        * title_text
-        * analysis_period
-        * values
-        * colors
 
         # TODO: Add this
         # * orientation_label_points
@@ -56,13 +185,15 @@ class WindRose(object):
         # * frequency_labels
         """
 
-    # TODO: Uncomment this.
-    # __slots__ = ('_direction_data_collection', '_speed_data_collection', '_center_point',
-    #              '_base_point', '_stack', '_radius', '_colored_mesh', '_container',
-    #              '_legend_parameters', '_chart_border')
+    __slots__ = ('_direction_data_collection', '_speed_data_collection',
+                 '_number_of_directions', '_legend_parameters', '_show_stack',
+                 '_show_zeros', '_base_point', '_base_point', '_radius',
+                 '_hist_bin_array', '_hist_bin_vectors', '_colored_mesh', '_border',
+                 '_orientation_lines', '_frequency_lines', '_container')
 
     def __init__(self, direction_data_collection, speed_data_collection,
-                 legend_parameters=None, stack=True, base_point=Point3D(), radius=1):
+                 number_of_directions=8, legend_parameters=None, base_point=Point3D(),
+                 radius=1, show_stack=True, show_zeros=True):
 
         """Initialize windrose plot."""
         # Check the input objects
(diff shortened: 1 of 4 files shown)


chriswmackey:
Thanks for getting this together, @saeranv . It's looking really good but, before I do a detailed review, I have a few borader comments:

1. I probably should have mentioned that there are a few use cases we have in mind for the wind rose beyond visualization. For example, If someone just needs to find the prevailing wind direction or someone wants to figure out what wind speeds should be used for a CFD study. For these simple types of applications, properties like the base_point, radius, legend_parameters, show_stack, show_zeros don't really have any meaning and so I wonder if they should be pulled out into arguments for methods on the class instead of being fixed read-only properties on the class. This is similar to [what you see over here on the sunpath where you only specify a radius and center point when you are requesting visual outputs like analemmas](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/sunpath.py#L462-L515)). I am thinking that this will help particularly with the case where we need to generate a colored mesh using data other than wind speed since this aligned data can be an input argument for this method (rather than us having to worry about how we assign this data as a property on the class).
2. You should take a look at the [recently added Compass class](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/compass.py#L13) since a lot of the comments I see in your code about adding orientation labels and orientation points are already addressed in this class. This means that you don't really need to have these properties on the wind rose class and users can just create a compass object if they need a visual of bounding circles.
3. Is there any way to fix the gap between frequency lines at a certain distance rather than being at the mercy of this being auto-generated based on the radius? I feel like not being able to fix this distance is going to make it difficult to compare certain wind roses.
4. I agree with moving those 3 functions to be on the DataCollection class. You can add them to the BaseCollection so that they are usable on all of the DataCollection classes.
5. Your [tests are failing in Python 2](https://travis-ci.org/github/ladybug-tools/ladybug/jobs/665125944?utm_medium=notification&utm_source=github_status). Please write them in a way that works for both Python 2 and Python 3 syntax.
6. In addition to docstrings, I think you will also need to add some more tests.

saeranv:
Thanks @chriswmackey, working on these fixes.

saeranv:
@chriswmackey 

This update should have everything resolved, except for the new tests, and moving those datacollection functions, which shouldn't hinder the detailed review. 

I rewrote the the visualization logic to get the fixed spacing working. So the fixed spacing will now scale the windrose automatically to max frequency of data w/ fixed gap sizeing if max frequency isn't explicitly set: 

image

Else it'll fix it w/ the fixed gap sizing.
image

And finally, the object initializes to just compute the binned wind analysis values:
```
w = WindRose(direction_values, analysis_values, 8)
w.data # list of list of analysis_values subdivided by direction values.
```

saeranv:
@chriswmackey 

I realized in the last version of this code, I was storing the editable visualization properties in such a way that modifying properties and computing the windrose mesh again wouldn't reflect the edited properties. So I made some changes to the way editable properties are stored, and the geometries are calculated so that they reflect changes to any editable properties. 

image

Also I caught a bug where the frequency lines weren't lining up to the value 'stacks'. Should be good for review now.

chriswmackey:
Thanks @saeranv . I will do a detailed review today. In the meantime, it would be good to add some more tests for the wind rose when you get the chance. You can see [in the coverage report here](https://coveralls.io/builds/29651954/source?filename=ladybug%2Fwindrose.py#L30) that the vast majority of the wind rose class that is currently not being tested.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+        return self._real_freq_max
+
+    @property
+    def container(self):
+        """GraphicContainer for the windrose mesh."""
+        # Create the graphic container from the inputs and associated legend properties
+        max_pt = Point3D(self._center_point2d.x + self._radius,
+                         self._center_point2d.y + self._radius, self.base_point.z)
+
+        return GraphicContainer(self.analysis_values, self.base_point, max_pt,
```
chriswmackey: I think is something you should store as a hidden property on the class and you should set it to None/regenerate it whenever the `LegendParameters` or `base_point` are set.  Then, you can have the `WindRose.legend` as a property returned from this graphic container like you see on the `HourlyPlot` module.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+    @property
+    def direction_values(self):
+        """The direction data values in this windrose plot."""
+        return self._direction_data_collection.values
+
+    @property
+    def data(self):
+        """A histogram of wind analysis values binned by the wind directions."""
+        return self._data
```
chriswmackey: This looks dangerous to me. Isn't the data a list? So a user could call this public property, append a bunch of stuff to this list and then the class has been mutated / messed up without any check. I would suggest either making this property hidden or using tuples instead of lists so that users can mutate the class.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+    @property
+    def frequency_lines(self):
+        """Frequency lines for windrose as Polygon2D lists."""
+        freqs = WindRose._ytick_radial_lines(self._bin_vectors,
+                                             self.legend_parameters.segment_count)
+        return [freq.scale(self._radius).move(self._center_point2d) for freq in freqs]
+
+    @property
+    def _center_point2d(self):
```
chriswmackey: This can go away if we just make the base point a `Point2D`

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+        """Compute center of windrose plot as Point2D."""
+        return Point2D(self.base_point.x + self._radius,
+                       self.base_point.y + self._radius)
+
+    @property
+    def _radius(self):
```
chriswmackey: Why isn't this public?  This is definitely something users will want in order to coordinate the wind rose with compasses or other compass-like graphics.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+        This number will be a integer multiple of the legend parameter segement count.
+        """
+        ytick_num = self.legend_parameters.segment_count
+        if self.show_zeros:
+            ytick_num += self._ytick_zero_inc
+        return self.frequency_spacing_distance * ytick_num
+
+    @property
+    def _wind_radius(self):
```
chriswmackey: I also don't understand why you would make this private.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+    @property
+    def _ytick_zero_inc(self):
+        """Returns how much yticks increasa with zeros"""
+        return ceil(self.zeros_per_bin / (self.real_freq_max + self.zeros_per_bin) *
+                    self.legend_parameters.segment_count)
+
+    @property
+    def _compass(self):
+        """Compass object for the windrose plot."""
```
chriswmackey: This should be public and then we can get rid of all the other properties that duplicate what the compass does.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+        """Radius of the windrose mesh."""
+        if self.show_zeros:
+            return (self.real_freq_max + self.zeros_per_bin) / \
+                (self.frequency_maximum + self.zeros_per_bin) * self._radius
+        else:
+            return self.real_freq_max / self.frequency_maximum * self._radius
+
+    @property
+    def _ytick_zero_inc(self):
+        """Returns how much yticks increasa with zeros"""
```
chriswmackey: Typo in docstring

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+
+def linspace(start, stop, num):
+    """Return evenly spaced numbers over calculated over the interval start, stop.
+
+    Equivalent to numpy.linspace(0, 5, 6) -> array([0., 1., 2., 3., 4., 5.])
+
+    Args:
+        start: Start interval index as integer
+        num: Number of divisions as integer
```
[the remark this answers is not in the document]
saeranv: Yes, my mistake the start and stop number can be a float as well.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+    Args:
+        values: Set of numerical data as a list.
+        bins: A monotonically increasing array of uniform-width bin edges, excluding the
+            rightmost edge.
+        hist_range: Optional parameter to define the lower and upper range of the
+            histogram as a tuple of numbers. If not provided the range is
+            ``(min(values), max(values)+1)``.
+        key: Optional function parameter to define key to bin values by. If not provided
+            the histogram will be binned by value item.
```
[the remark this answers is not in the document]
saeranv: The 'key' argument is an actual function. Similar to how you can pass functions to sort lists based on keys. 

i.e sorted(values, key = lambda k: k[0])

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+    Args:
+
+        direction_data_collection: A HourlyContinuousCollection or
+            HourlyDiscontinuousCollection of wind directions which will be used to
+            generate the windrose.
+        direction_data_collection: A HourlyContinuousCollection or
+            HourlyDiscontinuousCollection of wind speeds which will be used to
+            generate the windrose.
+        number_of_directions: The number of orientations to divided the wind rose
+            directions.
```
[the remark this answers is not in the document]
saeranv: Thanks! Yes, the docstrings here are out of date. Updated.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+        # Fixed properties for visualization that only need to be computed once
+        self._bin_vectors = None
+        self._zeros_per_bin = None
+        self._orientation_labels = None
+        self._real_freq_max = None
+
+    @property
+    def base_point(self):
+        """An optional Point3D used as the bottom left corner of the windrose geometry.
```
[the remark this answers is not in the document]
saeranv: @chriswmackey I originally tried this but the GraphicContainer class expects a Point3D for min and max_point. Should we change the GraphicContainer class or keep as is?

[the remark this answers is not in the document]
saeranv: Ref: Line 47 - 50 in graphic.py

> /app/ladybug/ladybug/graphic.py in __init__(self, values, min_point, max_point, legend_parameters, data_type, unit)
     46         # check the inputs
     47         assert isinstance(min_point, Point3D), \
---> 48             'min_point should be a ladybug Point3D. Got {}'.format(type(min_point))
     49         assert isinstance(max_point, Point3D), \
     50             'max_point should be a ladybug Point3D. Got {}'.format(type(max_point))

ladybug/windrose.py
saeranv: I'll make it into tuples. We need this exposed so people can access the histogram.

ladybug/windrose.py
saeranv: See question above: do we need to change GraphicContainer to accept a 2D?

ladybug/windrose.py
saeranv: Yeah we can expose this.

ladybug/windrose.py
saeranv: Exposed

ladybug/windrose.py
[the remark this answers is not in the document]
chriswmackey: The `GraphicContainer` is written that way because it has to deal with a wide variety of visualizations. For the wind rose, I think we should probably keep it as a Point2D publicly and, under the hood, we convert it to a Point3D with 0 for the z value.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+    def label_tick_size_factor(self, label_tick_size_factor):
+        """Set the size of the ticks for labels in the windrose plot."""
+        self._label_tick_size_factor = label_tick_size_factor
+
+    @property
+    def legend_parameters(self):
+        """The LegendParameters object for this plot.
+
+        Default is a legend parameter with minimum and maximum values of the
+        windrose analysis values. If show_stack is False, then these min and max
```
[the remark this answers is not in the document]
saeranv: I just realised I need to rewrite the histogram_data to show a nested, nested data structure (list of 'stack-intervals' nested within list of bins) to accurately show the show_stack values, since it's the average of the stack-intervals that represent the range of the windrose when show-stack = True, not the real min/max values. 

That way the values that get set into the Legend can be fixed as averaged values, and the user can modify other properties.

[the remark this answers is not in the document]
saeranv: I'll just delete the legend_parameters and let the legend property handle everything as discussed.

ladybug/windrose.py
```diff
@@ -0,0 +1,953 @@
...
+
+    @property
+    def legend(self):
+        """The Legend object for this plot"""
+        if self._legend is None:
+            self._legend = self._container._legend
+        return self._legend
+
+    @legend.setter
+    def legend(self, legend):
```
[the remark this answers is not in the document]
saeranv: Just to clarify, I think we can allow users to set the legend parameters, and any changes to that will propagate to the WindRose's Legend and GraphicContainer objects since Python passes pointers to complex data types. In theory we shouldn't have to manually regenerate the objects. It's a little confusing since we're counting on mutations to do all the work for us, but that's unavoidable either way.

ladybug/windrose.py
saeranv: I'll check this with some unit tests tomorrow, but I think we complex datatypes like objects will be mutated within other objects, and we don't need to set it to None/regenerate whenever we change LegendParameters/base_point. If I'm wrong I'll modify to None.

ladybug/windrose.py
saeranv: Since the basepoint, radius, and legend_parameters segment count all impact the way this object, we will have to set this property to None if any of those properties chagne. I'll add this as a note for future coders. As discussed, I think this might get messy as we scale up if we're mutating properties invisibly in other properties. I think the best way to handle this is just to cache these properties based on function arguements (memoization) which allows us to cache different instances of these classes based on changing properties without having to mutate it in other properties.

ladybug/windrose.py
saeranv: Nevermind. It looks like the radius is being used to intialize this object so my comment no longer applies (it's not just a object that sits within another object that chagnes, but the actual container class itself which will change.)

ladybug/windrose.py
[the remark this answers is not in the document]
saeranv: Nevermind. It looks like the radius is being used to intialize this object so my comment no longer applies (it's not just a object that sits within another object that chagnes, but the actual container class itself which will change.)

saeranv:
@chriswmackey here's the changes to the previous PR. I left some comments unresolved to continue conversation on some implementation details but essentially implemented everything you suggested. I think the only unresolved issue is whether we need to change GraphicContainer class to accept Point2Ds. 

Also apologies for still not having unit tests, I ran out of time today to get them completed, but can commit them tomorrow.

chriswmackey:
Thank you for the hustle, @saeranv . I will do a final pass over the code and respond to any unresolved comments as soon as I can. And no worries about the tests. We can get to them after you add the non-abridged objects to honeybee-schema.

chriswmackey:
Also, I see that you have a bunch of tests commented out there and hopefully we can just tweak those to get the tests that we need for CI.

saeranv:
@chriswmackey 

This PR includes the base_point change, new unit tests and the change the to the frequency binning so that the legend parameters reflect average value of each interval, not just min/max like I had previously. 

There are some bugs in the unit tests relating to angles which is why it's commented out, but I'll try and resolve it after the HB schema objects.

ladybug/windrose.py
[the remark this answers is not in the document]
chriswmackey: Yes. Sorry I should have clarified this. You can see [the Legend object also duplicates any legend parameters that are passed to it](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/legend.py#L116). So, this way, the original legend parameters are not mutated as defaults are set. In other words, if you want to edit the legend parameters assigned to a particular legend or graphic container, you have to get the legend_parameters on those objects themselves.

chriswmackey:
Thanks @saeranv . Make sure to also get some sleep!

ladybug/windrose.py
```diff
@@ -0,0 +1,761 @@
...
+
+    @property
+    def real_freq_max(self):
+        """The maximum frequency of the windrose data."""
+        if self._real_freq_max is None:
+            self._real_freq_max = max([len(d) for d in self.histogram_data])
+        return self._real_freq_max
+
+    @property
+    def _container(self):
```
chriswmackey, line 282: Can we make this property public? I don't think we are gaining anything by making it private but there are things that users will gain by being able to access it. For example, they can get the `value_colors` from the container that relate to the `colored_mesh`.

ladybug/windrose.py
```diff
@@ -0,0 +1,761 @@
...
+    def wind_radius(self):
+        """Get the radius of the windrose mesh."""
+        if self.show_zeros:
+            return (self.real_freq_max + self.zeros_per_bin) / \
+                (self.frequency_maximum + self.zeros_per_bin) * self.radius
+        else:
+            return self.real_freq_max / self.frequency_maximum * self.radius
+
+    @property
+    def _ytick_zero_inc(self):
```
chriswmackey, line 395: This is minor but you don't really need to use a `@property` decorator here if the method is already hidden. The major reason why you'd use property decorators in the first place is to signify to API users that something is read-only (or you have a special setter to check the input is it isn't read-only) but, when it's already hidden, the decorator isn't really doing anything for you except saving you from an extra `()` when you call it.

ladybug/windrose.py
```diff
@@ -0,0 +1,761 @@
...
+    def _compute_colored_mesh_array(hist_data, hist_data_stacked, bin_vecs, ytick_num,
+                                    min_radius, max_radius, show_stack):
+        """Compute a colored mesh from this object's historam.
+
+        Args:
+            regl_hist_data: Histogram of analysis values greater then zero as a list of
+                lists.
+            bin_vecs: Array of histogram bin edge vectors.
+            ytick_num: Number of Y-axis intervals.
+            show_stack: Boolean indicating if stacked histogram.
```
chriswmackey, line 698: I know this is a hidden method but you might want to update this docstring.

chriswmackey:
Actually, I didn't see the updated coverage report. Feel free to merge it once you address the small comments and be sure to squash it before you merge so that we don't add all 18 of you commits to the master commit log.

saeranv:
Merged!

PR #243, opened 2020-04-29
docs(datacollection): Fixing code-block error in docs.

saeranv:
Adding missing line that was causing error (hopefully).


saeranv:
@mostaphaRoudsari I thought I did that by linking the issue? Or is there something else I'm supposed to be doing?

image

PR #245, opened 2020-05-03
Added '__slots__'

mkrueger12:
Add `__slots__` to epw.py

### DIFF
--- a/ladybug/epw.py
+++ b/ladybug/epw.py
@@ -74,6 +74,11 @@ class EPW(object):
         * liquid_precipitation_quantity
         * sky_temperature
     """
+    __slots__ = ("_file_path", "_is_header_loaded", "_is_data_loaded", "_is_ip",
+                 "_data", "_heating_dict", "_cooling_dict", "_extremes_dict",
+                 "_extreme_hot_weeks", "_extreme_cold_weeks", "_typical_weeks",
+                 "_monthly_ground_temps", "_is_leap_year", "daylight_savings_start",
+                 "daylight_savings_end", "_num_of_fields")
 
     def __init__(self, file_path):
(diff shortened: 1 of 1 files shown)


mostaphaRoudsari:
Also I just check the tests and you are at least missing `_matadata`. You can see it for yourself here: https://travis-ci.org/github/ladybug-tools/ladybug/jobs/682683655

mkrueger12:
I have updated the file and corrected the syntax. Do I submit a new PR?

mostaphaRoudsari:
Thanks. You don't need to send a new PR. Every new commit to your `master` branch will update this PR automatically.

Thank you for the changes! The tests are still failing. You can run the tests locally using ` pytest ./tests` too. 😀

I see that you have also changed a number of dependencies in the file. Is that intentional? I don't think we need to change the imports for adding `__slots__`.

mkrueger12:
Sounds good. I will run the test locally. This is my first time doing this so thanks for the help. I'm not sure what may have caused the dependency sorting but it was unintentional.

mostaphaRoudsari:
> This is my first time doing this so thanks for the help.

I figured. no worries. We all started one day and went through a similar process. Let me know if you have any other questions. Will be happy to help. 😀

> I will run the test locally.

Sounds good. If you haven't already you may want to install the dependencies using `pip install -r dev-requirements.txt`

> I'm not sure what may have caused the dependency sorting but it was unintentional.

You might have an extension installed on your editor that does it for you on save.

mkrueger12:
epw_test.py runs without error locally. Thoughts?

mostaphaRoudsari:
Thank you @mkrueger12! Well done. The tests are passing. I added @chriswmackey to have another look and make the final call.

PR #247, opened 2020-05-04
perf(slots): Add '__slots__' to STAT object

mkrueger12:
Test succeeded locally.


mkrueger12:
perf(slots): Add '__slots__' to WEA object

mkrueger12:
Thanks @mostaphaRoudsari. I have updated my fork and will look into `feature branches` as well. Anything I need to do for the stat.py and wea.py commits right now?

PR #248, opened 2020-05-04
Add `__eq__` method to Wea

mostaphaRoudsari:
During the implementation of `SkyMatrix` I needed to test the equality of two Wea objects. In order to implement `__eq__` to `Wea` I had to implement it for other classes that are part of Wea object.

related #206


ladybug/_datacollectionbase.py
```diff
@@ -940,6 +940,18 @@ def is_mutable(self):
         """Boolean denoting whether the data collection is mutable."""
         return self._mutable
 
+    def __key(self):
+        return (
+            self.header, self.values, self.datetimes, self.validated_a_period,
+            self._enumeration
```
chriswmackey, line 946: You really don't want to check that the `_enumeration` is equal in order to tell if two collections are equal.  This `_enumeration` is only there to help with the conversion between immutable and mutable collections and you can see that it doesn't even appear in `__slots__`. It's just some data that needs to be auto-generated after loading the class but only in situations where you actually need to convert between mutable and immutable collections.

ladybug/_datacollectionbase.py
```diff
@@ -940,6 +940,18 @@ def is_mutable(self):
...
+        )
+
+    def __eq__(self, other):
+        return isinstance(other, BaseCollection) and self.__key() == other.__key()
```
chriswmackey, line 950: Unless you want to re-write this function each time it appears in the 8 classes that inherit from it, I would change this to:
```
return isinstance(other, self.__class__) and self.__key() == other.__key()
```
As it is right now, it could produce some odd results.

ladybug/_datacollectionbase.py
chriswmackey: Also, I would recommend overwriting this `__key` method for the `HourlyContinuousDataCollection`. Since the majority of data collections we use are of that type and there's no need to check that the `datetimes` and `validated_a_period `are equal on that class. This will generally be much faster for the common cases we use the data collection for.

ladybug/datatype/base.py
```diff
@@ -46,6 +46,13 @@ def __init__(self, name=None):
         """
         self._name = name
 
+    def __key(self):
+        return (
+            self._name, self._units, self._si_units, self._ip_units, self._min,
+            self._max, self._abbreviation, self._unit_descr, self._point_in_time,
+            self._cumulative, self._normalized_type, self._type_enumeration
```
chriswmackey, line 53: You definitely don't want to check that the `_type_enumeration` is equal for the same reason you don't want to check that the data collection enumeration isn't equal.

ladybug/datatype/base.py
```diff
@@ -298,6 +305,15 @@ def ToString(self):
         """Overwrite .NET ToString."""
         return self.__repr__()
 
+    def __hash__(self):
+        return hash(self.__key())
```
chriswmackey, line 309: This is going to fail for some data types since you aren't making sure that all of the elements of the `__key` are immutable. For one, the `_unit_descr` is a dictionary and you would need to sort it by keys and turn it into two tuples in order to be able to `__hash__` it. So I say that you should remove the `__hash__` method unless you are going to do it correctly.

ladybug/header.py
```diff
@@ -129,6 +129,18 @@ def ToString(self):
         """Overwrite .NET ToString."""
         return self.__repr__()
 
+    def __key(self):
+        return (self.data_type, self.unit, self.analysis_period, self.metadata)
+
+    def __hash__(self):
+        return hash(self.__key())
```
chriswmackey, line 136: This is also going to fail since you aren't hashing the `data_type`, the `analysis_period` in the key. Nor are you converting the `metadata` into an immutable format. So Just remove this for now unless you are want to implement it correctly.

mostaphaRoudsari:
Hi @chriswmackey, Thank you for the review. I addressed all the comments and took your advice for `__hash__` methods and removed them.

PR #249, opened 2020-05-07
feat(sun): Add a property for sun to get azimuth including north

chriswmackey:
I realized that this is needed for radiance skies.

### DIFF
--- a/ladybug/sunpath.py
+++ b/ladybug/sunpath.py
@@ -994,8 +994,19 @@ class Sun(object):
 
     @property
     def azimuth(self):
-        """Return solar azimuth in degrees."""
+        """Return solar azimuth in degrees.
+
+        This value is the same regardless of what the north_angle is.
+        """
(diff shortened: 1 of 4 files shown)


ladybug/sunpath.py
```diff
@@ -994,9 +994,20 @@ def altitude(self):
...
-        """Return solar azimuth in degrees."""
+        """Return solar azimuth in degrees.
+
+        This value is the same regardless of what the north_angle is.
+        """
         return self._azimuth
 
+    @property
+    def azimuth_from_y_axis(self):
+        """Return solar the azimuth in degrees with respect to the Y-axis of the scene.
```
mostaphaRoudsari, line 1005:
```suggestion
        """Return solar azimuth in degrees with respect to the Y-axis of the scene.
```

chriswmackey: Thank you.

PR #264, opened 2020-05-19
fix(wea): Simplify Wea dictionary schema

chriswmackey:
There were several redundant properties within the Wea schema, which was fine when all dictionaries were originating from the ladybug-core library. But, now that I have been documenting the Wea schema for radiance workflows, the schema really should be refactored to minimize the potential for conflicts and need for checks between these redundant properties.

I also added a check that the location assigned to the Wea is a Location object.

### DIFF
--- a/ladybug/wea.py
+++ b/ladybug/wea.py
@@ -60,18 +60,20 @@ class Wea(object):
         * timestep
         * is_leap_year
     """
-    __slots__ = ('_timestep', '_is_leap_year', 'location', 'metadata',
-                 '_direct_normal_irradiance', 'dhr', '_diffuse_horizontal_irradiance')
+    __slots__ = ('_timestep', '_is_leap_year', '_location', 'metadata',
+                 '_direct_normal_irradiance', '_diffuse_horizontal_irradiance')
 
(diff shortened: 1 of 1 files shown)


chriswmackey:
@mostaphaRoudsari ,

The references in that issue are so old that I can't understand what the problem is. Is it that the datetimes are still an annual timeseries but a few are not following the expected interval of the timestep?

If so, what you are asking for is a much deeper overhaul of the structure of the ladybug Wea Python object (not just the dictionary schema).  The HourlyContinuousDataCollections that are used to store the irradiance values on the Python object only work with continuous data at a fixed timestep. We would have to convert them to discontinuous data collections to accommodate a changing timestep.  And this will also make it much more time consuming to check that the datetimes of all of the data are aligned with one another.

Can we just merge this for now and do your big overhaul in another PR?

chriswmackey:
I can open the issue but, as I said in my response, I don't know about all of the broken references in the issue that you linked to. So I opened it but I think you will have to elaborate on what use cases actually have to be addressed:
https://github.com/ladybug-tools/ladybug/issues/265

PR #284, opened 2020-06-23
feat(windrose): Modifying windrose for GH component.

saeranv:
@chriswmackey 
cc @mostaphaRoudsari 

These are improvements to the wind rose module that the new GH component relies on. 

Changes:
- Adding more settable properties: north, different frequency controls
- Change frequency calculation to be another histogram calculation 
- Made defaults more explicit
- More detailed doc strings
- More tests
- General code clean-up/corrections

### DIFF
--- a/ladybug/windrose.py
+++ b/ladybug/windrose.py
@@ -1,9 +1,12 @@
 # coding=utf-8
 from __future__ import division
 
+import math
+
+from .color import ColorRange
 from .datacollection import HourlyContinuousCollection, HourlyDiscontinuousCollection
 from .graphic import GraphicContainer
-from .legend import LegendParameters
+from .legend import LegendParameters, LegendParametersCategorized
 from .compass import Compass
 
 from ladybug_geometry.geometry2d.pointvector import Point2D
@@ -11,8 +14,6 @@ from ladybug_geometry.geometry2d.line import LineSegment2D
 from ladybug_geometry.geometry2d.polygon import Polygon2D
 from ladybug_geometry.geometry2d.mesh import Mesh2D
 from ladybug_geometry.geometry3d.pointvector import Point3D
-
-import math
 
 # Simplify method names
 linspace = HourlyContinuousCollection.linspace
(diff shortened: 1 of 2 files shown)


ladybug/windrose.py
```diff
@@ -188,19 +232,37 @@ def legend_parameters(self):
...
 
+    @property
+    def north(self):
+        """North orientation for windrose by degrees in counterclockwise orienation.
+
+        This must be a number between -360 and 360. 0 is North, 90 is West and 270
+        is East.
+        """
+        if self._north is None:
+            self._north = self.DEFAULT_NORTH
```
chriswmackey, line 254: Why not just `return self.DEFAULT_NORTH`? Here you are loosing track of the fact that the user did not specify this value for `_north`. Or, if this information about whether the north was user-assigned is not important, why bother setting `_north` to `None` in the first place.

ladybug/windrose.py
```diff
@@ -188,19 +232,37 @@ def legend_parameters(self):
...
+        return self._north
+
+    @north.setter
+    def north(self, north):
+        assert isinstance(north, float), 'north must be a float representing degrees.' \
+            ' Got {}.'.format(north)
+
+        self._compass = None
+        self._container = None
+        self._north = north % 360.0
```
chriswmackey, line 264: Why do you divide by 360 here? Is the `_north` not in degrees?

saeranv: Yes this is in degrees. But it's not a slash, it's a modulo sign (%). It's just there to reset values > 360 back to 0.

ladybug/windrose.py
saeranv: I see your point, it doesn't have to be set to None in the first place. I'll set it to the default value in the __init__ function.

ladybug/windrose.py
chriswmackey: Ah, that's actually very elegant. Sorry that I forgot the modulo is the remainder of the division and not the result of the division itself.

chriswmackey:
Thanks @saeranv . I am going to merge this now.

PR #288, opened 2020-06-28
fix(windrose): Fix mesh size bug, and frequency_maximum to interval.

saeranv:
@chriswmackey 

This PR contains the bug fix for the windmesh scaling issue I mentioned in my previous PR. The windrose is now scaling correctly with and without the calmrose. 

Also I changed the way users control the maximum frequency from them setting maximum hours, to setting number of frequency lines in the windrose plot, which should be a lot more intuitive.

### DIFF
--- a/ladybug/windrose.py
+++ b/ladybug/windrose.py
@@ -58,7 +58,8 @@ class WindRose(object):
         * orientation_lines
         * frequency_lines
         * frequency_spacing_distance
-        * frequency_intervals
+        * frequency_intervals_compass
+        * frequency_intervals_mesh
         * frequency_maximum
         * frequency_hours
     """
@@ -193,7 +194,7 @@ class WindRose(object):
 
     @property
     def frequency_maximum(self):
-        """Get or set maximum frequency to represent in the windrose plot.
+        """Get maximum frequency to represent in the windrose plot.
 
         Default: number of items in largest windrose histogram bin.
         """
@@ -201,7 +202,7 @@ class WindRose(object):
 
     @property
     def frequency_hours(self):
-        """Number of hours to bin analysis values, per direction."""
+        """Get or set number of hours to bin analysis values, per direction."""
         if self._frequency_hours is None:
             self._frequency_hours = self.DEFAULT_FREQUENCY_HOURS
         return self._frequency_hours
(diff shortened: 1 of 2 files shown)


ladybug/windrose.py
```diff
@@ -222,8 +213,25 @@ def frequency_hours(self, frequency_hours):
         self._frequency_hours = int(frequency_hours)
 
     @property
-    def frequency_intervals(self):
-        return int(math.ceil(self.frequency_maximum / self.frequency_hours))
+    def frequency_intervals_compass(self):
+        """Calculate the number of intervals in the windrose compass."""
```
chriswmackey, line 217: This doesn't seem like the right type of docstring for a property. Property docstrings should start with "Get" or "Get or set" depending on whether they are set-able.

ladybug/windrose.py
```diff
@@ -222,8 +213,25 @@ def frequency_hours(self, frequency_hours):
...
+        if self._frequency_intervals_compass is None:
+            self._frequency_intervals_compass = self.frequency_intervals_mesh
+        return self._frequency_intervals_compass
+
+    @frequency_intervals_compass.setter
+    def frequency_intervals_compass(self, frequency_intervals_compass):
+        """Calculate the number of intervals in the windrose compass."""
```
chriswmackey, line 224: Same comment here.

ladybug/windrose.py
```diff
@@ -222,8 +213,25 @@ def frequency_hours(self, frequency_hours):
...
+        assert frequency_intervals_compass >= 1, 'The frequency_intervals_compass ' \
+            'must be greater then 0. Got: {}'.format(frequency_intervals_compass)
+        self._compass = None
+        self._container = None
+        self._frequency_intervals_compass = int(frequency_intervals_compass)
+
+    @property
+    def frequency_intervals_mesh(self):
+        """Calculate the number of intervals in the windrose mesh."""
```
chriswmackey, line 233: And here.

saeranv:
@chriswmackey 

Thanks, I realized that I really haven't been following that rule consistently so went through and fixed it in all my properties.

chriswmackey:
Great! Let's merge.

PR #289, opened 2020-06-29
fix(windrose): Remove rounding of bin lengths when averaging wind_data.

saeranv:
@chriswmackey 

Sorry Chris, I'm adding one more PR. Currently the lengths of the wind rose 'bins' are integer multiples of the frequency intervals, and I realized this shouldn't be the case when we don't show frequency intervals and just average all the data per bin.


chriswmackey:
It looks like the edits to the Grasshopper component shouldn't require any more edits here so I am going to merge.

PR #295, opened 2020-07-03
fix(windrose): Fix prevailing wind direction getter.

saeranv:
@chriswmackey 

Here's the fix. I should have made sure to write a test for it the first time.

### DIFF
--- a/ladybug/windrose.py
+++ b/ladybug/windrose.py
@@ -326,15 +326,33 @@ class WindRose(object):
 
     @property
     def prevailing_direction(self):
-        """Get the predominant direction of the wind values.
-
-        In case of ties, this property will return the first direction value in the
-        ordered direction values.
+        """Get a tuple of the predominant directions of the wind values.
         """
+
         if self._prevailing_direction is None:
-            dirvals = sorted(self.direction_values)
-            self._prevailing_direction = dirvals[dirvals.index(max(dirvals))]
-        return self._prevailing_direction % 360.0
+            # group by frequency
+            groups = {}
+            for dirval in self.direction_values:
+                dirval = int(dirval)
+                if dirval not in groups:
+                    groups[dirval] = 1
+                else:
+                    groups[dirval] += 1
+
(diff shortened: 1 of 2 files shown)


saeranv:
Fascinating, it looks like in Python 2.7 the dictionary iterables are in the opposite order of Python 3, which is causing a test to fail. Will change and recommit.

ladybug/windrose.py
```diff
@@ -326,15 +326,33 @@ def histogram_data(self):
...
-            self._prevailing_direction = dirvals[dirvals.index(max(dirvals))]
-        return self._prevailing_direction % 360.0
+            # group by frequency
+            groups = {}
+            for dirval in self.direction_values:
+                dirval = int(dirval)
+                if dirval not in groups:
+                    groups[dirval] = 1
+                else:
+                    groups[dirval] += 1
```
chriswmackey, line 340: I feel like this is really going to backfire on us if we ever have an EPW where the wind direction doesn't fall into 10-degree increments. Can we use the binned `_histogram_data` and the user-input `number_of_directions` for this case instead?

ladybug/windrose.py
```diff
@@ -326,15 +326,33 @@ def histogram_data(self):
...
+            max_freq = 0
+            for dirv, freq in groups.items():
+                if max_freq < freq:
+                    max_dirv = [dirv]  # Reset list
+                    max_freq = freq
+                elif max_freq == freq:
+                    max_dirv.append(dirv)
+
+            # set tuple
+            self._prevailing_direction = tuple(max_dirv)
```
chriswmackey, line 353: Also, why is this a tuple? Shouldn't it be a single value for the direction in degrees?

saeranv: That's to account for the edge case of ties (same frequency of values in two or more directions).

saeranv:
@chriswmackey  This change will return the direction based on the number of directions of the windrose.

Issue #265, opened 2020-05-19
Allow the editing of specific datetimes in the Wea object


chriswmackey:
Admittedly, I don't know exactly what needs to be supported here but @mostaphaRoudsari will detail this out more.  At the least, it seems that the the Wea object needs to support a changing timestep:

https://github.com/ladybug-tools/ladybug/pull/264
https://github.com/ladybug-tools/honeybee/issues/130

Maybe it also needs to support non-annual values?

mostaphaRoudsari:
Wea has not been the easiest object to deal with. I think source of most of these issues is because wea is being used for something that it is not designed to do.

The original intent was to create an **annual** weather tape which carries irradiance values for each hour of the year. Daysim was using this file to generate the annual sky for annual daylight studies. In this settings and to capture the cumulative nature of irradiance each time is shifted for 30 minutes so it represents the time between the two timesteps. Other than that Daysim had to do an adjustment for hours close to the horizon as they had issues for generating the Perez sky correctly. So the rule of hours being continuous with the same timestep was broken but it was fine for Daysim since the end to end approach is hard-coded and user cannot make any changes to it.

There was no option for sub-annual studies and originally there was no option for creating wea with smaller timesteps. As a result you code have map the values based on index not the real time. For example the 8th value would always represent Jan, 1, 8:00am regardless of it being 8:30, 8:25 or 8:15.

When timestep was introduced by keeping the file annual the same trick can be used, and this is what we are doing in the current wea. You can get the related hour based on the index as you always know which index is supposed to represent which hour of the year.  https://github.com/ladybug-tools/ladybug/blob/7ea7cf664d78f81d658aa0572befbf5b7c8b4996/ladybug/wea.py#L589

To add on top of the Daysim way of generating wea was one way. It would just generate the file and the concept of reading it back or modifying it was non-existent. That makes the life easy. It's always annual and the values will be reported based on their order. It was also never designed to return the values for a specific hoy or a sub-annual study. As a matter of fact Daysim would only give you the annual metrics. Honeybee was the tool that started to expose the hourly values. It was not meant to be looked at really.

[...]

PR #310, opened 2020-07-24
feat(wea): Add support for discontinuous data in Wea objects

chriswmackey:
This commit makes it possible to use hourly discontinuous data collections for Wea objects. It also adds a range of methods to produce filtered Weas based on certain criteria including:

* filter_by_pattern
* filter_by_analysis_period
* filter_by_hoys
* filter_by_moys
* filter_by_sun_up

I also fixed a bug in the previous performance improvement of AnalysisPeriod.__len__() that I added last week. And I added a from_start_end_datetime method to AnalysisPeriod that I am already making use of in this commit and I know I will use elsewhere. Lastly, I verified that the AnalysisPeriod.from_analysis_period method is not being used by any of the Grasshopper plugins and I removed it before any other developers feel the temptation to make use of it. It just feels like it's a liability the longer it remains in the code base and it's best to rip the band aid off now.

Lastly, I fixed a small bug in the data collection __key() property and improved / standardized a few of the docstrings that I came across.

There are a couple of Grasshopper components using the Wea class that I should update to align with this PR but, otherwise, this is all ready to merge.

Resolves https://github.com/ladybug-tools/ladybug/issues/265


ladybug/wea.py
```diff
@@ -547,60 +643,113 @@ def global_horizontal_irradiance(self):
...
             direct_horiz.append(dnr * math.sin(math.radians(sun.altitude)))
-        return HourlyContinuousCollection(header_dhr, direct_horiz)
+        return self._aligned_collection(header_dhr, direct_horiz)
 
-    @property
-    def is_leap_year(self):
-        """Return the timestep."""
-        return self._is_leap_year
+    def filter_by_pattern(self, pattern):
+        """Filter this Wea based on a list of booleans.
```
mostaphaRoudsari, line 668: Since we are returning a new Wea object I suggest to change the docstring to reflect that:

```suggestion
        """Create a new filtered Wea from this Wea based on a list of booleans.
```

mostaphaRoudsari: Similar comment about the next couple of methods.

ladybug/wea.py
```diff
@@ -753,89 +900,80 @@ def estimate_illuminance_components(self, dew_point):
...
-
-                lines.append(line)
+
+        # write values and datetimes
+        for dir_rad, dif_rad, dt in zip(self.direct_normal_irradiance,
+                                        self.diffuse_horizontal_irradiance,
+                                        self.datetimes):
+            line = "%d %d %.3f %d %d\n" \
+                % (dt.month, dt.day, dt.float_hour, dir_rad, dif_rad)
+            lines.append(line)
```
mostaphaRoudsari, line 955: This is much cleaner than what it used to be but I wonder how are you handling the 30 minutes shift for datetime? I see that the test file that is created looks correct but couldn't see where do you make the change in write method.

chriswmackey: The change for the datetimes happens in the datetimes property of the Wea:
https://github.com/ladybug-tools/ladybug/blob/master/ladybug/wea.py#L475-L482
You can see from where that code is that is how we were doing it in the past and I think it's a good solution. Admittedly, it means the datetimes property for the Wea might not be exactly as the datetimes are for the individual data collections. But this seems appropriate given that the Wea is following a different time standard from the rest of ladybug.

chriswmackey:
Hey @mostaphaRoudsari . I understand the questions you are raising but I think that adding half of the timestep minutes to the start of the series of datetimes creates a lot of work for us and, most of the time, I think it's probably not what people want. Intuitively, it's much easier for me to understand the following list of datetimes:
```
[00:00, 00:15, 00:30, 00:45, 01:00]
```
... than it is to understand this list of datetimes:
```
[00:08, 00:23, 00:38, 00:53, 01:08]
```
As you can see, it also gets into messy sub-minute details where we have to round the number of minutes since half of 15 minutes is 7.5 minutes.

Also, I think we can still easily align the results to Ladybug Tools occupancy schedules that use "time of beginning."  We had to do a similar mapping from EnergyPlus's "time until"and I think it's actually easier to map Ladybug Tools' "time of beginning" to the "average time" that the Wea is using.  We'll just continue to assume that an occupancy schedule from 9 AM to 5 PM includes 9:00 and includes 4:59 but excludes 5:00.

ladybug/wea.py
mostaphaRoudsari: Thank you for the clarification. This means that the file will be identical to the output of epw2wea command which is all we really need.

mostaphaRoudsari:
> Also, I think we can still easily align the results to Ladybug Tools occupancy schedules that use "time of beginning." We had to do a similar mapping from EnergyPlus's "time until"and I think it's actually easier to map Ladybug Tools' "time of beginning" to the "average time" that the Wea is using. We'll just continue to assume that an occupancy schedule from 9 AM to 5 PM includes 9:00 and includes 4:59 but excludes 5:00.

I'm not clear about this but I will reach out to you when we are re-writing the results parser for annual daylight simulation.

chriswmackey:
Thanks @mostaphaRoudsari . Thinking over your comments made me realize that there has been a small bug in the `from_epw` method when it is used at finer timesteps. We should make sure that the irradiance values from the EPW fall on the half hour and we just interpolate between these half-hours (rather than treating them like they fall on the hour mark as we do now). So I'm going to implement this fix and then merge soon after.

chriswmackey:
Never mind my previous comment. It seems the intelligence of shifting the data by a half hour is already handled by the `DataCollection.interoplate_to_timestep` method. I still added a new test for this case, though.

PR #487, opened 2021-04-06
docs(sphinx-conf): Fix CLI documentation process

santiagogaray:
- correctly handles commands without groups
- Check for group names inside cli\__init__.py

Latest CLI  doc. process update.


chriswmackey (approved):
Looks awesome. Thanks, @santiagogaray !

PR #538, opened 2021-12-17
fix(epw): add time-zone to metadata

devngc:


chriswmackey (approved):
Hey @devngc ,
I am cool with this change. Do you mind my asking what you need the time zone for before we merge? I am guessing that it is not for solar positions since you aren't also adding the latitude and longitude.

I just want to make sure there isn't some other easy way to get the time zone for your case before we merge this.

PR #539, opened 2021-12-21
fix(epw): Fixed a bug in the EPW.to_wea header

built-tools:
The term for "weather_data_file_units" is being written out as "weather_data_file_unit". This causes dctimestep to raise an error.

Before the fix: 
2021-12-21 15_49_15-D__Seattle wea - Notepad++

After the fix:
2021-12-21 15_49_36-D__Seattle wea - Notepad++


built-tools:
Additional notes: 
1. This is indeed a bug i.e. not trying to change something to make it work with Radiance.
2. The header has been correctly[ set for the Wea class.](https://github.com/ladybug-tools/ladybug/blob/master/ladybug/wea.py#L602-L609)

mostaphaRoudsari:
Hi @built-tools! Thank you for the fix. I'm merging it in.

PR #591, opened 2023-01-05
fix(sunpath): Ensure daylight savings is respected for sunrise/sunset

chriswmackey:


charliebrookerBH:
Hi @chriswmackey, hope you're well. 

I just opened the old script I made around this and it wasn't behaving as expected. 

I think that if is_daylight_savings, then it should be +1 to the times, rather than -1. Hopefully I've got that right, DST is a pain!!
image

Cheers,
Charlie

PR #606, opened 2023-11-30
fix(translate): Add option for epw input

mikkelkp:


ladybug/cli/translate.py
```diff
@@ -101,9 +102,14 @@ def wea_to_constant(wea_file, value, output_file):
...
     \b
     Args:
-        wea_file: Full path to .wea file.
+        wea_file: Full path to .wea file. This can also be an .epw file.
     """
     try:
+        try:
+            _wea_file = os.path.join(os.path.dirname(wea_file), 'epw_to_wea.wea')
+            wea_file = Wea.from_epw_file(wea_file).write(_wea_file)
+        except Exception:
```
chriswmackey, line 111: There are definitely much cleaner and more efficient ways to implement this rather than trying to treat the file like an EPW and then waiting for it to fail. This is particularly because supplying a Wea instead of an EPW is not an exceptional case but it's the most common one given our current usage. It's far better to just read the first word of the file to tell what type it is, which is always `place` in a Wea and is always `LOCATION` in an EPW.

I recommend the following, which is very similar to [how we check whether a given model file is a JSON or a binary Pkl](https://github.com/ladybug-tools/honeybee-core/blob/master/honeybee/model.py#L249-L251) or [how we check whether a given model file is a zipped radiance folder vs. a Honeybee model](https://github.com/ladybug-tools/honeybee-radiance/blob/master/honeybee_radiance/cli/translate.py#L86-L92):

```
with open(wea_file) as inf:
    first_word = inf.read(5)
is_wea = True if first_word == 'place' else False
```

This is what I would recommend doing for all of the cases where you're trying to accept either an EPW or Wea.

ladybug/cli/translate.py
```diff
@@ -101,9 +102,14 @@ def wea_to_constant(wea_file, value, output_file):
```
chriswmackey, line 102: Also, you should update the docstrings of the command to make it clear that the input can either be a Wea or an EPW.

chriswmackey (changes requested):
Thanks @mikkelkp .
There are a couple of changes that I would recommend before we merge.

chriswmackey (approved):
LGTM. Let's merge!

